PR #129, opened 2015-01-19
[PIR-1113] Email of HTML output from Interactive Report loses formatting

bmorrise:


bmorrise:
@mdamour1976 Could you look at this and merge it if it is acceptable?

PR #132, opened 2015-01-20
[SP-1295][PRD-5086] - When auto-submit is off, the report parameter/prom...

SergeyTravin:
...pt panel flashes/glass pane appears.

Made changes so that the glass pane is only shown with a dark background when the report is being updated (this is the most time-taking process).
When the glass pane is shown, but only the prompt is being refreshed, it is shown transparent, achieving a less disruptive user experience.

The incremental recreation of each of the prompt's CDF controls still significantly disrupts the user experience. It looks like a slide-in effect.

* Summarising, I made two changes:
** replaced all explicit calls to show/hide the glass pane's dojo element with calls to central functions having the same code; this change is conservative of the original behaviour - a refactoring;
** then (a behaviour changing change), the new show/hide methods add logic that additionally controls the glass pane's transparency.

Detailing the changes:
* In reportviewer-prompt.js
** Created two methods in the report-viewer prompt showGlassPane and hideGlassPane whose implementation calls show/hide on the dojo element.
** Replaced all existing calls to show/hide the dojo element with calls to showGlassPane and hideGlassPane.
** Moved the call to showGlassPane in getParameterDefinition to fetchParameterDefinition.
*** This location is more central: it accounts for other scenarios where fetchParameterDefinition is being called directly (in reportViewer.js), and thus required additional explicit calls to showGlassPane.
* In reportviewer.js
** In the same spirit of other existing code, the default prompt showGlassPane and hideGlassPane implementations are overridden with view-specific implementations. These implementations use details only known by the view (whether the report content is loading as well or not) to control the transparency of the glass pane, when showing/hiding.
* A new method was added, "reportContentUpdating", which uses the already existing "_updateReportTimeout" variable.
* Taking advantage that fetchParameterDefinition already calls showGlassPane, the existing call in _updateReportContent can be moved to the only branch of code which doesn't issue it: the _updateReportContentCore method. It is placed right after the update to "_updateReportTimeout" variable, so that showGlassPane can know that the glass pane must not be transparent.

Conflicts:
	package-res/reportviewer/reportviewer-prompt.js
	package-res/reportviewer/reportviewer.js

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -49,9 +49,6 @@ pen.define(['common-ui/util/util', 'common-ui/util/formatting'], function(util, 
         //    PromptPanel._ready ->
         //    
         panel.getParameterDefinition = function(promptPanel, callback) {
-          // Show glass pane when updating the prompt.
-          dijit.byId('glassPane').show();
-
           // promptPanel === panel
           this.fetchParameterDefinition(promptPanel, callback, /*promptMode*/'USERINPUT');
         }.bind(this);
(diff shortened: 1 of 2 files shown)


SergeyTravin:
@mdamour1976, It is backport for https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/85 Could you review it please?

SergeyTravin:
@mdamour1976, @pamval, @dkincade please comment/review.

PR #146, opened 2015-06-30
[BACKLOG-3310] - (Using BISERVER-6.0-SNAPSHOT) Error opening Interaive report

rfellows:
The content generator instances were hanging onto stale versions of member variables (from previous executions) and the incorrect information was being used (like request parameters and report definitions) that did not apply.


rfellows:
@rmansoor please review

PR #160, opened 2015-08-25
[BACKLOG-4038]

diogofscmariano:
- Moved the glass pane hide, allowing it to occur every time the submit ended, even if the prompts still need an input from the user

@pamval @krivera-pentaho @annapopova1 @rfellows @scottyaslan @plagoa please review
Dont merged it yet. Tomorrow i want to do more tests and check every unit test. But the implementation should be finished


diogofscmariano:
dependent on https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/486

PR #161, opened 2015-08-27
[BACKLOG-4265] Enable POST for report viewing

scottyaslan:


package-res/reportviewer/report.html
```diff
@@ -116,7 +116,10 @@
 
 <div id="reportArea" class="pentaho-transparent">
   <div id="reportPageOutline">
-    <iframe id="reportContent" frameborder="0"></iframe>
+    <form id="hiddenReportContentForm" method="post" target="reportContentIframe">
```
krivera-pentaho: Since we are already using jquery, is there a reason you are using a hidden form instead of using jquery to perform the post?
http://api.jquery.com/jquery.post/

scottyaslan: Absolutely! The reason is because of the iframe. There are things happening once the report is loaded etc. that trigger the functionality between PUC and that iframe. This approach leaves all of that alone, POST's the data as required and loads it in the existing iframe. Everyone listening for that or interacting with it behaves as before.

krivera-pentaho: What things are listening to the form you just added? I am not talking about the iframe, which is still present, but the form. The only thing that other elements would be listening to is when the iFrame is loaded, but not when the POST of the new form is complete.

Though it is technically correct, using a form to perform simple POSTs is a little antiquated (kind of like putting inline styles in an HTML tag instead of having CSS). And for manipulating/loading content into an iFrame from the result of the POST, you can do the solution that is provided in this article (http://stackoverflow.com/questions/8226307/set-iframe-innerhtml-without-loading-page-in-it-with-jquery) which is to get the iFrame and write the result of the POST to the contentDocument of the iFrame.

pentaho-nbaker:
We use This Form/Target technique in a lot of areas of the product already. Writing out to the DOM of an empty Iframe isn't as straight-forward as one would like to believe. <scripts> won't load synchronously, document.writes won't work, CSS can load in the wrong order. Form posting to an Iframe to get around the request length limit is a very valid technique.

PR #163, opened 2015-08-31
[BACKLOG-4038] - Review the prompting lifecycle

plagoa:
- Fixed small issue where the glass pane was preventing the user from changing his selection when Auto-Submit wasn't checked;


plagoa:
@diogofscmariano @annapopova1 @krivera-pentaho @scottyaslan @rfellows @pamval please review

krivera-pentaho:
Does this change need to be ported into PIR as well or is this only for report viewer?

PR #166, opened 2015-09-15
[BACKLOG-5149] - prpt Date Picker Prompt does not behave as expected …

plagoa:
…in PUC

- Added validation for timestamp values


plagoa:
@diogofscmariano @annapopova1 @krivera-pentaho @scottyaslan @rfellows @pamval please review

PR #177, opened 2015-11-10
PDB-1869 - Dashboard displays a blank PIR report if it contains a Pro…

Aliaksandr-Kastenka:
…mpt with option 'Allow multiple selections'

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
+++ b/src/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
@@ -106,13 +106,27 @@ public class ReportContentUtil {
 
   public static Object computeParameterValue( final ParameterContext parameterContext,
       final ParameterDefinitionEntry parameterDefinition, final Object value ) throws ReportProcessingException {
-    if ( value == null ) {
+      Object def = null;
+      if ( value == null ) {
       // there are still buggy report definitions out there ...
-      return null;
+      if ( parameterDefinition.getDefaultValue( parameterContext ) != null ) {
+        def = parameterDefinition.getDefaultValue( parameterContext );
+      } else {
+        return null;
+      }
     }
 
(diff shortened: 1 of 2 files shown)


Aliaksandr-Kastenka:
I closed previous pull-request https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/287
@pamval @diogofscmariano 
Please review. If it is ok, i will update pr with test.

src/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
```diff
@@ -106,13 +106,27 @@ public static ValidationResult applyInputsToReportParameters( final MasterReport
...
+      if ( parameterDefinition.getDefaultValue( parameterContext ) != null ) {
+        def = parameterDefinition.getDefaultValue( parameterContext );
+      } else {
+        return null;
+      }
     }
 
     final Class valueType = parameterDefinition.getValueType();
     final boolean allowMultiSelect = isAllowMultiSelect( parameterDefinition );
+
```
pamval: Isn't this too strong ? Shouldn't this be done only when we know the parameter is of type List ?
Otherwise, we're converting every string with a comma to an array list value which could be wrong.

I could be wrong but that's what it seems like it's being done here.

Aliaksandr-Kastenka:
@pamval 
When user simply opens prpti-report, this logic is in use:
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/master/src/com/pentaho/iadhoc/service/util/ThinParameterConverter.java#L128-L136
default value is setting by ThinParameterConverter
When user add prpti-report in dashboard this logic is not called, value is not set and this leads to the issue.

pamval:
ok, sounds sort of reasonable. @diogofscmariano  please process pull request

PR #192, opened 2015-12-21
BACKLOG-5887 - Dashboard cannot resolve parameter with a default value.

Aliaksandr-Kastenka:
@tmorgner @pamval 
Please review.
When server parse JSON and save all data into created XML-document it call toAttributeValue method
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/util/beans/ArrayValueConverter.java#L69
string with default values, separated by comma is builded.
when user open report in interactive reporting plugin this string is splitting to create an array
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/master/src/com/pentaho/iadhoc/service/util/ThinParameterConverter.java#L128-L136
similar logic is added here for the case when user add prpti-report into dashboard.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
+++ b/src/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
@@ -26,6 +26,7 @@ import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.Map;
@@ -46,6 +47,7 @@ import org.pentaho.reporting.engine.classic.core.parameters.ParameterContext;
 import org.pentaho.reporting.engine.classic.core.parameters.ParameterDefinitionEntry;
 import org.pentaho.reporting.engine.classic.core.parameters.ValidationMessage;
 import org.pentaho.reporting.engine.classic.core.parameters.ValidationResult;
+import org.pentaho.reporting.engine.classic.core.parameters.DefaultListParameter;
 import org.pentaho.reporting.engine.classic.core.util.ReportParameterValues;
 import org.pentaho.reporting.engine.classic.core.util.beans.BeanException;
 import org.pentaho.reporting.engine.classic.core.util.beans.ConverterRegistry;
@@ -53,6 +55,7 @@ import org.pentaho.reporting.engine.classic.core.util.beans.ValueConverter;
 import org.pentaho.reporting.libraries.base.util.StringUtils;
 import org.pentaho.reporting.libraries.resourceloader.ResourceException;
 import org.pentaho.reporting.platform.plugin.messages.Messages;
+import org.pentaho.reporting.libraries.base.util.CSVTokenizer;
 
 public class ReportContentUtil {
 
@@ -86,8 +89,23 @@ public class ReportContentUtil {
       for ( final ParameterDefinitionEntry param : params ) {
         final String paramName = param.getName();
         try {
+          Object value = inputs.get( paramName );
+          if ( value == null && ReportContentUtil.isAllowMultiSelect( param ) ) {
+            if ( param.getDefaultValue( context ) != null ) {
+              Object defaultValue = param.getDefaultValue( context );
+              if ( defaultValue instanceof String && param instanceof DefaultListParameter ) {
+                CSVTokenizer csvt = new CSVTokenizer( (String) defaultValue );
+                ArrayList<String> result = new ArrayList<String>();
+                while ( csvt.hasMoreTokens() ) {
+                  final String el = csvt.nextToken();
+                  result.add( el );
+                }
+                ( (DefaultListParameter) param ).setDefaultValue( result.toArray( new String[ result.size() ] ) );
+              }
+            }
+          }
           final Object computedParameter =
-              ReportContentUtil.computeParameterValue( context, param, inputs.get( paramName ) );
+              ReportContentUtil.computeParameterValue( context, param, value );
           parameterValues.put( param.getName(), computedParameter );
           if ( log.isInfoEnabled() ) {
             log.info( Messages.getInstance().getString( "ReportPlugin.infoParameterValues", //$NON-NLS-1$
--- a/test-src/org/pentaho/reporting/platform/plugin/ReportContentUtilTest.java
+++ b/test-src/org/pentaho/reporting/platform/plugin/ReportContentUtilTest.java
@@ -1,0 +1,60 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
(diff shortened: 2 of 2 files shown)


tmorgner:
My question still stands: Why does PIR generate a parameter/prompt that needs this sort of hack, when the same report generated in PRD does not? Shouldnt there be a fix for PIR to make sure that PIR sets up the parameter correctly?

Aliaksandr-Kastenka:
@tmorgner 
"Why does PIR generate a parameter/prompt that needs this sort of hack"
I don't know why it was implemented this way, the code which saves default values into XML parameter attribute as a string:
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/writer/DataDefinitionFileWriter.java#L194-L200
 was brought with inital check-in by Nick Baker and did not changed from that time.
@pentaho-nbaker maybe you will shed some light on this.

tmorgner:
Well, that is NOT the problem. This change there is about XML writing. As you know, XML is a text based format, so at some point attribute values must be converted into strings. If you were to look at the parsing part, you'd see that the parser then passes the attribute value back into the ConverterRegistry to convert the XML attribute value back into a proper array.

The code you highlighted simply checks whether the returnned string is null or empty to prevent the writing of an redundant xml-attribute. It has, once again, nothing to do with the problem at hand.

Given that the reporting engine and the reporting plugin code has not changed in these parts since 5.0 or before, can we all agree that there may be a good chance that a fix concerns the dashboard-plugin itself? 

The dashboard gets its parameter info from the parameter-xml that the reporting plugin generates. So your first port of call should be to validate that this is correct. Hint: If the report-viewer works and recognizes the parameters and their default values, chances are that this part is working.

Your next port of call should then be the dashboard designer. check that the widget-prpt.js (refreshParameters method) is able to parse the XML returned by the server. Can it do that? Does it see the correct values?  

If so, then now trace how these values are passed back and forth within the dashboard designer. Any fix should now be limited to the dashboard designers code, I would suppose.

Aliaksandr-Kastenka:
@tmorgner 
I suppose this may be the root cause:
"If you were to look at the parsing part, you'd see that the parser then passes the attribute value back into the ConverterRegistry to convert the XML attribute value back into a proper array."
The problem is that it's not working like it's expected, when user adds report(for example i got report with two selected prompt values: "Cancelled" and "In Process") into dashboard and method startParsing gets executed at this point:
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/data/AbstractParameterReadHandler.java#L76
type equals "java.lang.String", so org.pentaho.reporting.engine.classic.core.util.beans.StringValueConverter is in use, instead of ArrayValueConverter and 
defaultValue will be "Cancelled,In Process" instead of array.

type came from JSON, in case of this type: "[Ljava.lang.String;" - it parsed correctly, i've tried to change it, but not succeed:
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/287

In other words, if i understood correctly at this point(line 76) type will never be array. That is the problem and should be changed.

tmorgner:
Again, you are seeming to mix the XML that we store on disk with what PIR/Dashboard sends over the wire. They are NOT directly connected. The PRPT file format is fixed, and PIR/Dashboards will NEVER EVER use the XML directly, it always parses the PRPT file to get a MasterReport object before doing any work. So if something is buggy, it has to do with the code that works off the MasterReport object or the JSON translation. So fix the PIR code, leave the reporting engine alone! 

As for JSON parsing: It is probably safe to say that the valueType can be left at "java.lang.String". The code should be looking at both the valueType and the multiSelect flag. If multiSelect is true, you are dealing with a multi-selection parameter and the default value can have multiple entries. 

Now if I look at the ThinParameterConverter class in PIR, I see some very suspicious code on how parameters with arrays are handled. The section commented with "unwrap single value array" for instance is wrong there. if you are dealing with a multi-selection parameter, the system needs an array of values, even if that array has only one element. Same goes for the "reinflate serialized value" comment right after that. Why would there be a need for that? 

So for PIR please go and do a proper investigation on why the JSON may contain CSV string instead of using proper arrays as default values. 

However, the PIR case may be a separate issue and you may additionally run into a PIR bug here.

So to narror it down and to solve the dashboarding bug: Test the issue with a properly defined report that has been defined in PRD, and that works in the report-designer. Then validate that multi-selection values from *that* report work and if not, fix the dashboard designer to make it work.

And then, only when you validated it with a known good sample, can you start bringing in PIR reports and potentially fix them too. 

But right now I feel you are trying to cheaply fix symptoms instead of researching and understanding the root cause of the problem.

Aliaksandr-Kastenka:
@tmorgner 
Sorry to bother you again, I investigated attached report(PDB-1869.prpi) - after publishing it is opens in PUC correctly, adding such report in dashboard also works correctly, i here is what i found:

report (in report designer)
when prompt with multi selection is created - DefaultListParameter created with
allowMultiSelection=true, valueType=**class [Ljava.lang.String;**
https://github.com/pentaho/pentaho-reporting/blob/master/designer/report-designer/src/org/pentaho/reporting/designer/core/editor/parameters/ParameterDialog.java#L1038-L1042
when user save report, at this point(writeListSelectionParameter):
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/writer/DataDefinitionFileWriter.java#L246
valueType=class [Ljava.lang.String;

report (in interactive reporting plugin)
when prompt with multi selection is created  - DefaultListParameter created with
allowMultiSelection=true, valueType=**class java.lang.String**
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/master/src/com/pentaho/iadhoc/service/util/ThinParameterConverter.java#L106-L108
when user save report, at this point(writeListSelectionParameter):
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/writer/DataDefinitionFileWriter.java#L246
valueType=class java.lang.String

After that when user open report(saved in report designer),
at this point(startParsing):
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/data/AbstractParameterReadHandler.java#L76
type=class [Ljava.lang.String;, ArrayValueConverter is in use, defaultValue is **array**

when user open report(saved in interactive reporting plugin),
at this point(startParsing):
https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/parser/bundle/data/AbstractParameterReadHandler.java#L76
type=class java.lang.String, StringValueConverter is in use, defaultValue is **String**
that is why "the JSON may contain CSV string instead of using proper arrays as default values".

If fix should be applied in PIR, I propose to edit ThinParameterConverter and make it similar to
https://github.com/pentaho/pentaho-reporting/blob/master/designer/report-designer/src/org/pentaho/reporting/designer/core/editor/parameters/ParameterDialog.java#L1038-L1042
so i need to know your opinion, thank you in advance.

tmorgner:
Yes, please fix PIR so that it handles mutli-selection parameters correctly. Dont forget to prove that your fix works via a unit-test. You can test whether your fix works by directly testing the properties and behaviours of the ParameterDefinition object returned by the ThinParameterConverter.

PR #194, opened 2016-01-21
[PIR-1163] Fixed the execution of main query in PIR when preparing to…

annapopova1:
… show schedulers parameters

Details:
The report viewer sends two server requests to get parameters XML. The first request gets initial XML and the second request gets additionally information about paging (accepted-page, page-count, paginate) and include it to parameters XML.
For scheduler paging is not necessary and we can avoid the second request.
For this purpose I use a special variable in report viewer "inSchedulerDialog". The variable is true when scheduler dialog is opened and false when it is closed. As a result the second request is sent when scheduler dialog is closed.

@diogofscmariano @krivera-pentaho @pamval @plagoa @jvelasques @afrjorge please review


krivera-pentaho:
Looks good to me, but I don't know enough about this bug to merge it.

PR #198, opened 2016-02-23
Ewok0

EgorZhuk:
Iteration 0 features changes.


pamval:
Guys, take notice that the commit message should be [<jira case>] - message
So the jira number should be written inside square brackets: [PRD-5021] and not PRD-5021

EgorZhuk:
@pamval  Ok, should I rewrite the history or just follow this rule in future?

pamval:
we either need to rewrite it now or when we merge to master. I'm fine with one or the other.

PR #199, opened 2016-02-23
Ewok0

EgorZhuk:
Changes for reporting team iteration 0. PRD-5021 and BISERVER-1225.
@tmorgner


tmorgner:
Can you please split this into two pull requests? One for the two PRD-5021 cases and one for the rest of the commits. The "merge" commit at the end is strange, it only modifies files that are not even touched by the PRD-5021 commits. It seems that this "merge" is a hidden bug fix and should be marked as such.

tmorgner:
General: In the Pentaho-platform and its plugins, the naming convention is that interfaces are prefixed with an upper-case I (basically the dot-net practice). Thus the interfaces in the async package should be named "IAsyncReportState" and so on. 

Second: PentahoAsyncExecutor should have an interface with its public methods defined there. Create an IPentahoAsyncExecutor interfce. Then in spring, use the simple classname a bean-id (ie "IPentahoAsyncExecutor"). You then allow OEMs to implement their own strategy if needed, and follow good Pentaho coding practice.

src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -0,0 +1,193 @@
...
+  private boolean isSuportAsync = true;
+
+  public JobManager( boolean isSuportAsync ) {
+    if ( !isSuportAsync ) {
+      logger.info( "JobManager initialization: async mode marked as disabled." );
+    }
+    this.isSuportAsync = isSuportAsync;
+  }
+
+  @GET public Response getEcho() {
```
tmorgner: Is this used? If not, then please remove that.

src/org/pentaho/reporting/platform/plugin/staging/AsyncJobFileStagingHandler.java
```diff
@@ -0,0 +1,189 @@
...
+
+  private static final Log logger = LogFactory.getLog( AsyncJobFileStagingHandler.class );
+
+  private TrackingOutputStream fileTrackingStream;
+  // package private for testing purpose
+  File tmpFile;
+
+  private static String stringParentDir = ".";
+
+  public AsyncJobFileStagingHandler( OutputStream outputStream, IPentahoSession userSession, String stringParentDir ) throws IOException {
```
tmorgner: If you dont use the output stream, then dont pass it in. The only time this class is instantiated is in a manual constructor call - so delete that parameter. 

Initialize() will fail if stringParentDir is null. This variable is only ever set in the constrcutor. So do that check here, not in initialize(), and thus fail early!

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
```diff
@@ -0,0 +1,128 @@
...
+
+  private ILogger logger;
+
+  private SimpleReportingComponent reportComponent;
+  private AsyncJobFileStagingHandler handler;
+  private String url;
+
+  private String userSession = "";
+  private String userName = "";
+  private String insnaceId = "";
```
tmorgner: Typo: should be "instanceId".

src/org/pentaho/reporting/platform/plugin/staging/AsyncJobFileStagingHandler.java
```diff
@@ -0,0 +1,189 @@
...
+  public InputStream getStagingContent() throws FileNotFoundException {
+    return new StagingInputStream( new FileInputStream( tmpFile ) );
+  }
+
+  @Override
+  public void close() {
+    this.closeQuietly();
+  }
+
+  @Override
```
tmorgner: Actually, you dont need the TrackingStream at all. Just ask your tmp-file for its file length. This way you also wont hold on to the streams for longer than necessary.

EgorZhuk:
Will split into two requests.

PR #201, opened 2016-02-29
BISERVER-1225 - async report executions

prakapenka:
Added functionality to async report executions,
Junit test added.
fixed build.xml file to be able to really install plugin on tomcat instance.


tmorgner:
Please remove 
org.pentaho.reporting.platform.plugin.async.AsyncReportStatusListener#status

This variable is not safe - you cannot store every single report-event. Remember the last one, forget about any previous event. @see the email I sent for the last pull request

> 
(2)	As for the queue in AsyncReportStatusListener: Never put code for a potential debugging session into production code! First, if there is a performance problem, you’d probably be running the report in isolation at the engine level. If there is a need to collect performance data, you’d write it into a log. In no case would you hold on to every single state change for every user and every report on every system ever run. 

There is still an "insnaceId" in the PentahoAsyncReportExecution class - the parameter of the "forInstanceId" has the same typo.

package-res/plugin.spring.xml
```diff
@@ -66,4 +69,17 @@
         class="org.pentaho.reporting.platform.plugin.cache.PluginCacheManagerImpl" scope="singleton">
     <constructor-arg ref="plugin-cache-strategy"/>
   </bean>
+
+  <!-- thread pool executor -->
+  <bean id="reporting-async-thread-pool" class="org.pentaho.reporting.platform.plugin.async.PentahoAsyncExecutor" scope="singleton" >
```
tmorgner: The bean ID should be IPentahoAsyncExecutor. This is the standard approach in Pentaho's spring files and we should not deviate from that standard.

prakapenka:
Dear Thomas, all functioanlity related to updates of listener will come immediately after this part will be merged, I will remove ability (or someone-else) to store events in collection. According to status updates - as you mention -  org.pentaho.reporting.platform.plugin.async.AsyncReportStatusListener#status
not sure we are really need to have a strict sync on get/set methods here - but if you insist I can implement 'compareAndSet' approach (or just volatile) - instead of regular set(...). But currently from my opinion we don't really need it as the only side effect of it could be one exessive request for processing status since (...->FINISHED) write was still not propagated. FINISHED is the last one state.

tmorgner:
Please make it a syncronized call. The Java memory model (@see JSR-133) states that changes to a memory location may not be seen by another thread (changes are not written, and cached values are not updated) until you hit a syncronization point. So both the set and get methods should be synchronized or all variables used for the status info should be volatile. (Both synchronize and access to a volatile variable will trigger a memory refresh and make all threads see the same values).

The overhead of synchronized is low, and the action happens infrequently enough to not be a problem - and it makes sure we have deterministic behaviour here.

prakapenka:
made listener async now

lucboudreau:
@tmorgner For the record, synch operations, albeit low overhead, do trigger a full flush of CPU caches. Even though this particular code won't see much of a hit, other pieces of code running in parallel would. They are to be used as little as possible.

The correct solution is to use AtomicReference, or other classes of the Atomic variety.

tmorgner:
Hmm, synchronize does not flush *all* CPU caches. AFAIK On a X86, for example, it uses the "mfence" or similar operations, which then serialize the load and store operations for that thread context/CPU. All other threads/CPUs happily continue their work unaffected from the locking. 

That is also the reason why you have to put serialize around both the read and write operations to guarantee correctness. 

Thus only the two threads that share that lock will be affected. AtomicReferences are good for lock-free algorithms, but they only free you from syncronized blocks if you deal with primitive types or certified read-only objects (objects where all member variables are final).

PR #204, opened 2016-03-03
1st page feature

EgorZhuk:
1st page backend part ( disabled by property before the UI task is completed )
Includes file system cache locking - needs a review.
@tmorgner


package-res/reportviewer/messages/messages.properties
```diff
@@ -21,4 +21,10 @@ SessionExpired=Session Expired
 SessionExpiredComment=Your session has expired. Please reload this page once you have logged in again.
 
 ErrorParsingParamXmlMessage=Error parsing parameter information.
-FatalErrorTitle=Fatal Error
\ No newline at end of file
+FatalErrorTitle=Fatal Error
+
+# Async processing report Activity name
+async_computing_layout_title=Computing layout
```
tmorgner: For these key names, please follow the CamelCase convention the other keys use. Consistency is a good thing.

package-res/reportviewer/messages/messages_de.properties
```diff
@@ -22,3 +22,9 @@ SessionExpiredComment=Ihre Sitzung ist abgelaufen. Bitte melden Sie sich erneut
 
 ErrorParsingParamXmlMessage=Fehler beim Analysieren der Parameterinformation.
 FatalErrorTitle=Schwerwiegender Fehler
```
tmorgner: Dont provide any translations unless you speak the language. Google translate is rather bad at that task, and the German here is nothing you'd want to see in a professional product ;)

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
```diff
@@ -21,25 +21,31 @@
...
+  public static final String COMPUTING_LAYOUT = "async_computing_layout_title";
+  public static final String PRECOMPUTING_VALUES = "async_precomputing_values_title";
+  public static final String PAGINATING = "async_paginating_title";
+  public static final String GENERATING_CONTENT = "async_generating_content_title";
+
   private volatile String path;
   private volatile UUID uuid;
   private volatile AsyncExecutionStatus status = AsyncExecutionStatus.QUEUED;
   private volatile int progress = 0;
+  private volatile int page = 0;
```
tmorgner: I didnt catch that the last time - but when you use synchronized, you dont need volatile. Synchronize blocks will already guarantee correctness. So its either volatile, or synchronized, but never both.

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
```diff
@@ -21,25 +21,31 @@
...
+  private volatile String activity;
+  private volatile int row = 0;
+  private volatile boolean firstPageMode = false;
+  private volatile String mimeType;
```
tmorgner: The member variables mimetype, path and uuid can be made final.

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
```diff
@@ -71,37 +77,125 @@ public synchronized int getProgress() {
...
   @Override
-  public synchronized void reportProcessingUpdate( ReportProgressEvent event ) {
-    // will be implemented in another commit
+  public synchronized void reportProcessingUpdate( final ReportProgressEvent event ) {
+    final int activity = event.getActivity();
+    if ( firstPageMode && ReportProgressEvent.GENERATING_CONTENT == activity && page > 0 ) {
+      //First page is ready here
+      this.status = AsyncExecutionStatus.CONTENT_AVAILABLE;
+    }
+    this.activity = getActivityCode( activity );
```
tmorgner: Those 4 lines are identical in all methods - refactor into common method?

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
```diff
@@ -71,37 +77,125 @@ public synchronized int getProgress() {
...
+    this.status = AsyncExecutionStatus.FINISHED;
+  }
+
+  public synchronized void setFirstPageMode( final boolean firstPageMode ) {
+    this.firstPageMode = firstPageMode;
   }
 
   @Override
-  public String toString() {
+  public synchronized IAsyncReportState clone() {
```
tmorgner: clone is back? Please do NOT use the clone() name unless you intending to uphold the contract *and* it makes sense to how you intend to use the method. Show your intend of how and where you use the method. 

You use the method to get a "ReportState", so please name it "getReportState". The fact that you make a copy of this instance is an implementation detail - a caller who just sees the public interface of your class should not need to know the details.

src/org/pentaho/reporting/platform/plugin/async/IAsyncReportState.java
```diff
@@ -29,6 +29,24 @@
...
+   * @return Page is currently being processed
+   */
+  int getPage();
+
+  /**
+   *
+   * @return Row is currently being processed
+   */
+  int getRow();
+
```
tmorgner: Instead of the row, which is useless for showing a progress bar, compute a percentage value that you can give to the GUI. This way the progress bar can go from empty to full as the report processes.

This means you need to take both row and max-row into account. Also remember: We have multiple passes, so a percentage based purely on result = row / maxRow will go from 0 to 100 and back to zero on each pass.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
```diff
@@ -51,50 +51,82 @@ public void setCachePath( final String cachePath ) {
...
       if ( !file.exists() ) {
         file.createNewFile();
       }
-
-      fout = new FileOutputStream( file );
-      oos = new ObjectOutputStream( fout );
-      oos.writeObject( value );
-      oos.close();
-      fout.close();
+      //closable resources
```
tmorgner: That code is work in progress, right? (The curly braces count dont match up). But yes, the fileoutput-stream should be closed via a finally block.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
```diff
@@ -51,50 +51,82 @@ public void setCachePath( final String cachePath ) {
...
+      unlock( filePath );
     }
-
     return true;
+
+  }
+
+  private void lock( final String filePath ) {
+    final File lock = new File( filePath + ".lock" );
+    while ( lock.exists() ) {
```
tmorgner: What's wrong with using file-locks. I mean real ones.

http://www.linuxtopia.org/online_books/programming_books/thinking_in_java/TIJ314_030.htm

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -71,6 +76,12 @@
   @Override
   public int paginate( final MasterReport report, final int yieldRate )
     throws ReportProcessingException, IOException, ContentIOException {
+    final ExtendedConfiguration config = ClassicEngineBoot.getInstance().getExtendedConfig();
+    final boolean firstPageMode =
+      config.getBoolProperty( "org.pentaho.reporting.platform.plugin.output.FirstPageMode" );
+    if ( firstPageMode ) {
```
tmorgner: No, dont do that. 

Change the report-viewer to not ask for paginated results (renderMode=XML vs renderMode=PARAMETER). When in async-mode, the report-viewer should never send parameter-xml requests with the renderMode=XML mode. Instead, compute and return the total number of pages as part of the status request, and then update the paginationControl in the report-viewer with those values.

This then renders your change here unnecessary as "paginate" will never be called on any output-handler.

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -161,31 +178,87 @@ private IReportContent produceCacheablePages( final MasterReport report, final i
...
+        cacheListener = new ReportProgressListener() {
+
+          @Override public void reportProcessingStarted( final ReportProgressEvent reportProgressEvent ) {
+            //ignore
           }
-        }
+
+          @Override public void reportProcessingUpdate( final ReportProgressEvent reportProgressEvent ) {
+            if ( reportProgressEvent.getActivity() == ReportProgressEvent.GENERATING_CONTENT
+              && reportProgressEvent.getPage() == 1 ) {
```
tmorgner: This feature is not "firstPage" mode as in return the page with number 1. The user requests a specific page via the "accept-page" parameter, and *that* is what the user wants. 

So when you hit the page specified by "acceptedPage" *then* you can return the content of that page. Always returning the first page will break the user's expectations.

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -161,31 +178,87 @@ private IReportContent produceCacheablePages( final MasterReport report, final i
     final PageableHtmlOutputProcessor outputProcessor = (PageableHtmlOutputProcessor) proc.getOutputProcessor();
     outputProcessor.setFlowSelector( new DisplayAllFlowSelector() );
 
-    final Map<Integer, byte[]> pages = new HashMap<>();
+    final ExtendedConfiguration config = ClassicEngineBoot.getInstance().getExtendedConfig();
+    final boolean firstPageMode =
+      config.getBoolProperty( "org.pentaho.reporting.platform.plugin.output.FirstPageMode" );
+
+    //Async listener
+    final ReportProgressListener listener = ReportListenerThreadHolder.getListener();
```
tmorgner: Listeners are informed in the order in which you add them. I'd add the cache-listener first, so that you can be sure that the content is written before you update the status. When you update the status, an eager user can try to fetch the content before it has been written. You can avoid that race condition by just having the correct order here.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
EgorZhuk: It's Java 7 try with resources

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
EgorZhuk: File locks are held on behalf of the entire Java virtual machine. They are not suitable for controlling access to a file by multiple threads within the same virtual machine.

https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileLock.html

@tmorgner

src/org/pentaho/reporting/platform/plugin/async/IAsyncReportState.java
EgorZhuk: `this.progress = (int) ReportProgressEvent.computePercentageComplete( event, true );`

It's already there. I will keep row in json as far as it is present on [mockup ](http://jira.pentaho.com/secure/attachment/64522/biserver-1225%20ux.png)

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
EgorZhuk: @tmorgner it means we can use FileLocks for locking between processes but not between threads in one process

tmorgner: And your "keep a file open for writing" allow two processes to write the same file? I dont think so.

if you follow my directions and use the user-id as one segment, your lock will only lock between the user. And if you name the lock like "${key}.lock" then your lock will be covering only the particular entry you are writing. This is as fine grained as it gets.

And remember: You need the user as ID as thanks to row-level security, the same hash can represent a report that produces different data depending on who executes it.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
tmorgner: Ahh, yeah. The curly brace and round braces look awfully the same on githubs font.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
tmorgner: I see.

EgorZhuk:
```
private FileLock getFileLock( final RandomAccessFile randomAccessFile ) throws IOException {
    FileLock lock = null;
    while ( lock == null ) {
      try {
        lock = randomAccessFile.getChannel().tryLock();
      } catch ( final OverlappingFileLockException ignored ) {
      }
      if ( lock == null ) {
        try {
          Thread.sleep( 100 );
        } catch ( final InterruptedException ignored ) {
        }
      }
    }
    return lock;
  }
```

@tmorgner This one will work for threads

EgorZhuk:
@tmorgner 
I've fixed most of issues.
The main problem is file backend lock/synchronization:
- FileLock doesn't work the way we need
- Creating .lock files is not a good idea too
- So I used lower garanty - at least users don't affect each other
Now I changed code to sync on session objects - it is equal to syncronized methods in session-scoped bean. I didn't manage to get session beans working in plugin - niether aop scoped-proxy nor explicitly calling to PentahoSystem.get() worked for me.
If you know what am I doing wrong - please point me to the right direction - it will be cleaner to make the bean session-scoped and sync entire methods.

If we still desire separate locks for each file - then I think it will be great idea to use more controllable backend.

EgorZhuk:
Applied proposed approach with Map @tmorgner

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
```diff
@@ -42,6 +45,7 @@
   private static final Log logger = LogFactory.getLog( FileSystemCacheBackend.class );
   public static final String REPLACEMENT = "_";
   public static final String SLASHES = "[/\\\\]+";
+  private final Map<List<String>, Object> syncMap = new HashMap<>();
```
tmorgner: Access to that map is not thread-safe. You synchronize once you have the lock, but the work of looking up the lock is not threadsafe. Use concurrent-hashmap here? Then you can write the updates in a way that is thread-safe in your "getLock(..)" method. A "putIfAbsent" call will make that operation safe and fast.

EgorZhuk:
@tmorgner  fixed. If now everything is ok - please let me know but don't merge, I will squash the commits to clean up the history

tmorgner:
As there is a newer pull request covering the same changes, lets close this one.

PR #205, opened 2016-03-04
[BACKLOG-6876] RV: Setup PromptAPI

annapopova1:


diogofscmariano:
no unit tests are needed on this case, unit tests will be build based on the other calls that we are going to change related with BACKLOG-6961

PR #207, opened 2016-03-04
[BACKLOG-6898] - Re-implement PromptPanel#getString function

plagoa:
- Removed function getString override


plagoa:
@pentaho/orlandocalisbon please review.

PR #216, opened 2016-03-09
1 st page feature clean history

EgorZhuk:
@tmorgner same as #204 but with clean history


src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
```diff
@@ -42,30 +42,31 @@ public void forInstanceId( String instanceId ) {
...
-
-    if ( reportComponent.execute() ) {
-      listener.setStatus( AsyncExecutionStatus.FINISHED );
+      if ( reportComponent.execute() ) {
 
-      long end = System.currentTimeMillis();
-      AuditHelper.audit( userSession, userSession, url, getClass().getName(), getClass()
+        final long end = System.currentTimeMillis();
+        AuditHelper.audit( userSession, userSession, url, getClass().getName(), getClass()
           .getName(), MessageTypes.FAILED, instanceId, "", ( (float) ( end - start ) / 1000 ), null );
```
tmorgner: This should be MessageTypes.SUCCESS

src/org/pentaho/reporting/platform/plugin/cache/DeleteOldOnAccessCache.java
```diff
@@ -81,18 +92,32 @@ private void cleanUp() {
...
+            backend.purge( compositeKey );
+            logger.debug( "Purged long-term cache: " + key );
+          }
         }
       }
     }
     logger.debug( "Finished periodical cache eviction" );
   }
+
+
```
tmorgner: Check key == null, or risk crashes. Session.getName() can be null if the user is not authenticated (ie you are anonymous).

src/org/pentaho/reporting/platform/plugin/cache/DeleteOldOnAccessCache.java
```diff
@@ -81,18 +92,32 @@ private void cleanUp() {
...
+  private String createKey( final String key ) {
+    try {
+      final MessageDigest md = MessageDigest.getInstance( "SHA-256" );
+      md.update( key.getBytes() );
+      final byte[] digest = md.digest();
+      return new String( Base64.encode( digest ) );
+    } catch ( final Exception b ) {
+      return null;
+    }
```
tmorgner: Returning <null> might not be a good idea. Throw an error and let the request fail (as we have no meaningful way of recovering here). 

In addition to that, a <null> element as a segment in the key for the storage backend is not handled and would cause an NPE. However, I cannot see a safe case where <null> should ever be allowed as segment id.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
```diff
@@ -51,64 +54,89 @@ public void setCachePath( final String cachePath ) {
...
   public boolean write( final List<String> key, final Serializable value ) {
-    final File file = new File( cachePath + StringUtils.join( cleanKey( key ), File.separator ) );
-
-    final ObjectOutputStream oos;
-    final FileOutputStream fout;
-    try {
-      file.getParentFile().mkdirs();
-      if ( !file.exists() ) {
-        file.createNewFile();
+    final List<String> cleanKey = cleanKey( key );
```
tmorgner: Can we rename the "cleanKey" method to "sanitizeKeySegments"? Clean-Key reads as if you remove content (ie "List#clear()"). Sanitize (as in "sanitizing input" ) is a more correct description of what the method does.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
```diff
@@ -51,64 +54,89 @@ public void setCachePath( final String cachePath ) {
...
+      return (Serializable) result;
+    }
+  }
+
+  /**
+   * Returns an object for read/write synchronization
+   * @param key compound key
+   * @return lock object
+   */
+  private synchronized Object getLock( final List<String> key ) {
```
tmorgner: No need to sync, no need to check if the "lock" returned is null. 

Your "putIfAbsent" call will never return null. If the map does not contain the lock, it will add the object you have as second parameter to the map and then return it to you. If there is already an lock in the map, it will return that, and will discard your "new Object()".

EgorZhuk: @tmorgner unfortunately not, from javadoc:

> Returns:
> the previous value associated with the specified key, or null if there was no mapping for the key

It's wierd.

tmorgner:
The Spring.xml still uses non-standard keys for its beans. Please make sure all cache-related beans follow the standard pentaho pattern. The standard pattern is: If there is an Interface my.package.IComponent , the Spring ID should be the simple class name ( here: IComponent)

Therefore key currently named as "plugin-cache-strategy" should read "IReportContentCache". Same for ICacheBackend and IReportContentCache.

src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackend.java
tmorgner: Nope, JavaDoc states otherwise. putIfAbsent is all you need:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentMap.html#putIfAbsent%28K,%20V%29

From that method's documentation:

If the specified key is not already associated with a value, associate it with the given value. This is equivalent to

   if (!map.containsKey(key))
       return map.put(key, value);
   else
       return map.get(key);

EgorZhuk: `import java.util.concurrent.ConcurrentHashMap;

public class Main {

    public static void main(String[] args) {
        ConcurrentHashMap<String, String> map = new ConcurrentHashMap<>();
        System.out.println(map.putIfAbsent("1", "2"));
        System.out.println(map.putIfAbsent("1", "3"));
        System.out.println(map.putIfAbsent("1", "4"));
    }
}
`

It will print null, 2, 2

@tmorgner

tmorgner: Damn it, what did I read last night then. OK, in that case, there is no point in using the concurrent hash map. If it does not safe us from having to lock down the object, we can use a normal hashmap and synchronize as normal.

EgorZhuk:
@tmorgner  additionally we will need changes in subfloor.xml to compile Java 8

EgorZhuk:
@tmorgner I rebased it with latest EWOK

Should I change versions in subfloor.xml?

`  <property name="javac.source" value="8" description="Provide source compatibility with specified release"/>
  <property name="javac.target" value="8" description="Generate class files for specific VM version"/>`

tmorgner:
Just one tiny detail:
Your IAsyncReportListener now leaks the fact that its main implementation is also a IAsyncReportState. That is 100% unnecessary. 

It is ok to directly reference the AsyncReportStatusListener implementation in the PentahoAsyncExecutor. In that class, you directly created the instance via "new ..", so you know what type the object has, and there is no way anyone can pass you a different implementation in. Also: The fact that you use an "AsyncReportStatusListener" is never mentioned in the public interface, so it is a well-hidden implementation detail. 

In PentahoAsyncExecutor, keep the map as "Map<CompositeKey, AsyncReportStatusListener> listeners". Then your "IAsyncReportListener" interface does not need to be a "IAsyncReportState". The users of the listener (responsible for attaching the listener to the report-process and for feeding state data in) are different from users of IAsyncReportState (reading the state data).

EgorZhuk:
@tmorgner fixed

EgorZhuk:
Thank you, Thomas.
@tmorgner

PR #218, opened 2016-03-09
[BACKLOG-6760]

diogofscmariano:
- Reviewed prompt panel instances across report viewer objects
  - Added getters and setter to access these properties properly
  - Removed all the instances used as arguments and used the class atributes


diogofscmariano:

diogofscmariano:
The change related with the BACKLOG-6760 about the setParameter api will follow this one, just doing some clean up to help all the prompt api calls we need to change and make sure everyone works over the same code base

krivera-pentaho:
Looks good to me, unless we need to "use strict"

PR #220, opened 2016-03-10
[BACKLOG-6961] - Adding a package.json to set the appropriate test-js…

jvelasques:
… dependencies' versions.

 - Working around a karma bug with loading the same file multiple times, asserting that all files in test-js are amd modules.


jvelasques:
I don't know if we'll want to have non-amd test utils, so I can remove those commented lines.
package.json inspired by the one in common-ui.

jvelasques:
@diogofscmariano went ahead and removed the commented lines

PR #222, opened 2016-03-11
[BACKLOG-6760] Fixed some issues of prompt usages

annapopova1:
Checked Report Viewer on biserver-ee-7.0-QAT-37 and found that the plugin works incorrectly.
Fixed next issues:
- panel() - is not a function
- incorrect 'this' usage in callback
- can't remove promptPanel argument for submit and getParameterDefinition functions because this behavior is related with PromptPanel implementation (uses promptPanel argument)

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -449,13 +449,13 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
           }
 
           pc.registerPageNumberChangeCallback(function(pageNumber) {
-            this.pageChanged(this.reportPrompt.panel, pageNumber);
+            this.pageChanged(pageNumber);
           }.bind(this));
         },
 
         pageChanged: function(pageNumber) {
           this._setAcceptedPage('' + (pageNumber - 1));
-          this.reportPrompt.panel.submit({isPageChange: true});
+          this.reportPrompt.panel.submit(this.reportPrompt.panel, {isPageChange: true});
         },
 
         togglePromptPanel: function() {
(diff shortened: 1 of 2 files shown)


diogofscmariano:
Thanks for the fixes anna :+1:

package-res/reportviewer/reportviewer.js
```diff
@@ -449,13 +449,13 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
           pc.registerPageNumberChangeCallback(function(pageNumber) {
-            this.pageChanged(this.reportPrompt.panel, pageNumber);
+            this.pageChanged(pageNumber);
           }.bind(this));
         },
 
         pageChanged: function(pageNumber) {
           this._setAcceptedPage('' + (pageNumber - 1));
-          this.reportPrompt.panel.submit({isPageChange: true});
+          this.reportPrompt.panel.submit(this.reportPrompt.panel, {isPageChange: true});
```
diogofscmariano: Why do we need to have the panel again as argument?

annapopova1: See https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/components/SubmitPromptComponent.js#L116
It uses _submit -> and set promptPanel argument to submit function:
https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/PromptPanel.js#L746
As a result without promptPanel arg in Report Viewer we loose options...

diogofscmariano: Ok. I agree. Lets put it there again and later when we review the submit we can review that. Thanks for that anna

PR #225, opened 2016-03-14
[BISERVER-1225] New json fields to match mockup

EgorZhuk:
Hello,
On mockup http://jira.pentaho.com/secure/attachment/64522/biserver-1225%20ux.png
we have rows / totalRows and pages / totalPages labels.
Appropriate fields added to json.
We may consider moving totalPages to ReportProgressEvent and implementing mutation logic in AbstractReportProcessor.
@tmorgner


tmorgner:
Agree: Providing the total pages via the progress-event makes sense - especially as we provide the current page in it. 

I merge this change anyway, as the code here does the job. I'd say handle that progress-even change in a new improvement backlog-case, so that it can be solved via the normal sprint process.

EgorZhuk:
BACKLOG-7159 created

tmorgner:
Hmm, I thought I had hit the merge button. Apparently not. Evil me.

PR #227, opened 2016-03-14
[BISERVER-1225] As an end user when executing a report run, I want fe…

Aliaksandr-Kastenka:
…edback on what is happening with the request(feedback screen updated according to the ux notes)

@tmorgner 
Please review.


tmorgner:
Is there a reason you hard-code styles via the style-attribute when we have a perfectly usable css file sitting in the project?

tmorgner:
I also noticed that the dialog text is not localizable. Please put the label-strings into the message-bundle and get the localized text from there.

Aliaksandr-Kastenka:
@tmorgner 
Please take a look.

PR #234, opened 2016-03-15
[BACKLOG-6961] Fixed setting renderMode for callbacks.

annapopova1:
1. Fixed wrong behavior with render mode.
Steps:
 
 - open Inventory List report
 - uncheck Auto-Submit
 - select Line -> Motorcycles
 - see query in console. Should be request with renderMode: PARAMETER but now we have renderMode: XML
 
2. Also removed unnecessary condition for inSchedulerDialog.

@pentaho/orlandocalisbon please review and merge

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -15,9 +15,9 @@
 * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
 */
 
-define(['common-ui/util/util', 'pentaho/common/Messages', "dijit/registry", 'common-ui/prompting/parameters/ParameterXmlParser', "common-ui/prompting/api/PromptingAPI", "common-ui/jquery-clean"],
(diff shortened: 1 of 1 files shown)


package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -94,8 +94,6 @@ define(['common-ui/util/util', 'pentaho/common/Messages', "dijit/registry", 'com
             if (initFlag) {
               this._createPromptPanelFetchCallback(paramDefn);
```
jvelasques: @pentaho/orlandocalisbon could we remove the need for this initFlag, using something like [_.once](http://underscorejs.org/#once) ?

annapopova1: probably need to try this.. I have not familiarized with _.once

annapopova1:
@pentaho/orlandocalisbon Applied _.once instead of init flag (thanks @jvelasques).

jvelasques:
@pentaho/orlandocalisbon looks good to me

PR #236, opened 2016-03-15
[PRD-5599] As an end user when running a paginated html report output…

Aliaksandr-Kastenka:
…, I want the first page of the report to appear once it's generated and not have to wait for the rest of the pages to be generated

@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -24,6 +24,7 @@ define(['common-ui/util/util', 'pentaho/common/Messages', "dijit/registry", 'com
     return logged({
       // The current prompt mode
       mode: 'INITIAL',
+      _isAsync: null,
 
       /**
        * Gets the prompt api instance
@@ -78,7 +79,7 @@ define(['common-ui/util/util', 'pentaho/common/Messages', "dijit/registry", 'com
             //
             // [PIR-1163] Used 'inSchedulerDialog' variable to make sure that the second request is not sent if it's scheduler dialog.
             // Because the scheduler needs only parameters without full XML.
-            if ( (typeof inSchedulerDialog === "undefined" || !inSchedulerDialog) && this.mode === 'INITIAL' && paramDefn.allowAutoSubmit()) {
+            if ( (typeof inSchedulerDialog === "undefined" || !inSchedulerDialog) && this.mode === 'INITIAL' && paramDefn.allowAutoSubmit() && !this._isAsync) {
               this.fetchParameterDefinition(paramDefnCallback.bind(this), 'MANUAL');
               return;
             }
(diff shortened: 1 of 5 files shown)


EgorZhuk:
> This will start a new Job. This is most likely NOT what you want here. If first-page-mode is enabled and the current job has signaled that the content for the first page is available, then you already know that you can read the content via the "/content" handler. So why another job start?

Report content is returned as future[InputStream] and will not be available until the whole report is fifnished. When we spawn a new job we get requested page from cache while the first job is taking care of finishing the whole report.

> The whole "if (isFirstPage)" and this else block makes no sense. The server will not send a "CONTENT_AVAILABLE" status if the first-page-mode is not enabled.

It's true.

> This if-statement makes no sense here. This code would only make sense if the first-page mode was limited to the first page. But it is not. Dont try to guess the state from the report-progress-event, let the CahingHTMLOutput update that status directly....

Agree, but  in this case we should also prevent switching it back to 'WORKING' when already was 'CONTENT_AVAILABLE'

> Dont query the content-entities to find out the page-number. The ReportProgressEvent has a "page" property that you can query directly. It will tell you the page the report is processing.

Will unit test it and fix.

> If you see an acceptedPage of less than zero, then your JavaScript client is still invoking the parameter-handler with renderMode=XML. In async-mode this is no longer needed as you will receive the total-page-number via the status from the job-manager.

It seems that 'accepted-page' param is missed if we don't send 'parameter' request with XML render mode. So, may be it should be fixed on client side.

Thomas, thank you very much for the review.

@tmorgner  @Aliaksandr-Kastenka

tmorgner:
If you send the parameter request as "renderMode=PARAMETER", you should not even call the CachingHTMLReportOutput, as no pagination should be done. If you have used the  "renderMode=PARAMETER" mode and ended here, then you have found a bug.

EgorZhuk:
@tmorgner Everything is ok with render mode PARAMETER the issue is in reportjob call itself, below I listed request parameters.

```
ts:1458113350487
output-target:table/html;page-mode=page
accepted-page:-1
showParameters:true
renderMode:REPORT
htmlProportionalWidth:false
1458113354121:1458113354121
```

@Aliaksandr-Kastenka

Aliaksandr-Kastenka:
The whole "if (isFirstPage)" and this else block makes no sense. The server will not send a "CONTENT_AVAILABLE" status if the first-page-mode is not enabled.

isFirstPage variable renamed to isFirstContAvStatus

tmorgner:
@EgorZhuk Thank you for investigating. 

As for:


>> This will start a new Job. This is most likely NOT what you want here. If first-page-mode is enabled and the current job has signaled that the content for the first page is available, then you already know that you can read the content via the "/content" handler. So why another job start?

> Report content is returned as future[InputStream] and will not be available until the whole report is fifnished. When we spawn a new job we get requested page from cache while the first job is taking care of finishing the whole report.

Yeah, I forgot we deal with futures in the backend and that those things hang on to their results 'til the bitter end. 

@Aliaksandr-Kastenka thank you for renaming "isFirstPage" - this cleaner name now indicates the purpose of the variable in the process and explains to the reader why you check it in the if-block. You could make it even more clear by moving "isFirstContAvStatus = false" directly after the "if". This way, anyone who reads the if immediately sees that this block is meant to be executed only once.

tmorgner:
>  the issue is in reportjob call itself

When you start the report-viewer for the first time, your page-control has no data. Therefore you have no current page selected by the user. Test whether accepted-page is -1 before sending the Ajax request to solve that. 

Later calls will reuse whatever page the user has selected.

There is another corner case we have to handle here: If the selected page is later than the total numbers of pages generated, the server should auto-correct that and return the last page of the report instead. 

This way changing parameters keeps you on the same page for as long as possible, but does not end you on an invalid page if the new parameter set returns less pages than before.

Aliaksandr-Kastenka:
@tmorgner @EgorZhuk
Please take a look.

tmorgner:
Looks good. Thank you for documenting the "acceptedPage + 2" reasoning as well.

PR #240, opened 2016-03-16
[BACKLOG-6863] - RV: Review autoSubmit state access

plagoa:
- Updated RV code to use the API prompting
- Added function to the report prompt to get state properties using the API.
- Updated the code to use this new function
- Added unit tests


tmorgner:
@pentaho/orlandocalisbon

tmorgner:
I think it looks good. Its a straight-forward change.

PR #243, opened 2016-03-17
Fix [BACKLOG-7081], refactoring to use AsyncExecutor through interfaces, comments added.

prakapenka:
Fix of ability to run unit test for Idea (don't interrupt parent thread and minimize side effects), refactoring to use interfaces when submitting async executions, fix of info message instead of interruption stack-trace since interruption is not an error.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
+++ b/src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
@@ -1,8 +1,11 @@
 package org.pentaho.reporting.platform.plugin.async;
 
 import org.apache.commons.io.input.NullInputStream;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.pentaho.platform.engine.core.audit.AuditHelper;
 import org.pentaho.platform.engine.core.audit.MessageTypes;
+import org.pentaho.reporting.engine.classic.core.ReportInterruptedException;
 import org.pentaho.reporting.platform.plugin.SimpleReportingComponent;
 import org.pentaho.reporting.platform.plugin.staging.AsyncJobFileStagingHandler;
 
@@ -11,6 +14,8 @@ import java.util.concurrent.FutureTask;
 import java.util.concurrent.RunnableFuture;
 
 public class PentahoAsyncReportExecution implements IAsyncReportExecution<InputStream> {
+
+  private static final Log logger = LogFactory.getLog( PentahoAsyncReportExecution.class );
 
   private SimpleReportingComponent reportComponent;
   private AsyncJobFileStagingHandler handler;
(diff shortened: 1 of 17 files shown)


prakapenka:
@tmorgner  please review.

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
```diff
@@ -45,10 +45,6 @@ public InputStream call() throws Exception {
 
       if ( reportComponent.execute() ) {
 
-        final long end = System.currentTimeMillis();
-        AuditHelper.audit( userSession, userSession, url, getClass().getName(), getClass()
-          .getName(), MessageTypes.INSTANCE_END, instanceId, "", ( (float) ( end - start ) / 1000 ), null );
-
         listener.setStatus( AsyncExecutionStatus.FINISHED );
```
tmorgner: What is the reason that we we no longer log the Audit information for the success of the execution?

prakapenka: Evil rebase made me do it.

prakapenka:
I'll fix and reopen

tmorgner:
Please split that pull request into two:

One for BACKLOG-6988, which is fine and can be merged.
One for BACKLOG-7081, for which you do a lot of refactoring that now once again merges the IAsyncListener with the IAsyncState. 

The listener interface should not reference the state.  As I explained to Yahor on his change (which you basically reverse), those two interfaces have different and clearly distinct use-cases and just because your implementation happens to put both on the same class does not justify merging these interfaces.

The only time you actually need the state from the listener is in your tests. And for that you dont need to pollute the public interface, you can simply cast your listener to its implementation and then access a public or package-protected getter for it. 

You then go on a large refactoring to the IAsyncReportExecutor interface, apparently wanting to get that from the Pentaho-ObjectFactory. 

First: If you make refactorings, keep them in their own commit so that they can be reviewed separately.

The Interface here should not have the "forSession" or "forUser" methods. Remove them, they are implementation specific. If you wanted to use the PentahoObjectFactory then you'd have to create a IAsyncReportExecutorFactory, which then can receive your session, instance and all the information that currently goes into the constructor. But the IAsyncReportExecutor interface itself is the wrong place for that.

Again: If you want to make that change, fine, but do it in a separate commit and preferably a separate pull request. 

As for the interruption fix: How does the reporting engine get informed that the job has been cancelled now? Yes, you set the status to Cancelled, but I cannot see a single Thread#interrupt() in your code. So  the reporting engine itself will never know that you want to end the processing early. 

In the JIRA case, I simply asked that you clear the interrupted flag when you receive a ReportInterruptedException. The whole refactoring just masks  the fact that this code no longer cancels the report at all. And without signaling an interrupt, it is no wonder that your test works.

And to top it up, you then dont even use IAsyncReportExecution as interface, so 

So: Please dont make these refactorings, just do the simple fix.

prakapenka:
If necessary to 'simply cast from interface to implementation' - classic example of improper interface use-case.

I need a state when return execution status or generating report as servlet output stream. IAsyncState is a read only object that represent a state of report generation. That can be listener, that can be any object that represent state. Since we have only one implementation - that is listener. 

But if we are prefer not to pollute public interface, and prefer to pollute exact code with interface to implementations casts I can implement it.

I'll take a look on PentahoObjectFactory.

As of interruption fix: thread will be interrupted on super.cancel() - which will invoke Future.cancel - see FutureTask.cancel(boolean ...) for reference. And report continue to do a cancel. The side effect was that having a thread, having a Future in this thread, and calling a method on a future cancel we seems used to recieve a link to current thread, instead of thread encapsulated in future. So we used to set interrupted flag for current thread, and for thread encapsulated for Future.

Will try to break commits and make rebase again.

PR #245, opened 2016-03-17
[BACKLOG-6988] - put info message instead of stacktrace in case of interruption

prakapenka:
catch and info in case of ReportInterruptedException


prakapenka:
@tmorgner please review

PR #247, opened 2016-03-17
[CLEAN-UP] - refactor IPentahoAsyncExecutor to be able to submit callable by interface

prakapenka:
- refactor IPentahoAsyncExecutor to be able to submit callable by interface
- remove redundant forSession forInstanceId methods form interface.
- move instance id to a constructor as optional (possible null) parametr

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/async/IAsyncReportListener.java
+++ b/src/org/pentaho/reporting/platform/plugin/async/IAsyncReportListener.java
@@ -2,7 +2,7 @@ package org.pentaho.reporting.platform.plugin.async;
 
 import org.pentaho.reporting.engine.classic.core.event.ReportProgressListener;
 
-public interface IAsyncReportListener extends ReportProgressListener {
+public interface IAsyncReportListener extends ReportProgressListener, IAsyncReportState {
 
   void setStatus( AsyncExecutionStatus status );
 
--- a/src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
+++ b/src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
@@ -85,9 +85,7 @@ public class BackgroundJobReportContentGenerator extends ParameterContentGenerat
     reportComponent.setOutputStream( handler.getStagingOutputStream() );
 
     PentahoAsyncReportExecution
-        asyncExec = new PentahoAsyncReportExecution( path, reportComponent, handler );
-    asyncExec.forSession( userSession.getId() );
-    asyncExec.forInstanceId( instanceId );
+        asyncExec = new PentahoAsyncReportExecution( path, reportComponent, handler, instanceId );
 
     PentahoAsyncExecutor executor =
       PentahoSystem.get( PentahoAsyncExecutor.class, PentahoAsyncExecutor.BEAN_NAME, null );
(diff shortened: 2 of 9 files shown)


prakapenka:
@tmorgner please review

tmorgner:
Please squash those two commits into one. After all, the first commit is not correct when standalone and thus has no right to exist as own entry in our commit-log. Squash, and the whole scope of work becomes clear.

src/org/pentaho/reporting/platform/plugin/async/IAsyncReportListener.java
```diff
@@ -2,7 +2,7 @@
 
 import org.pentaho.reporting.engine.classic.core.event.ReportProgressListener;
 
-public interface IAsyncReportListener extends ReportProgressListener {
+public interface IAsyncReportListener extends ReportProgressListener, IAsyncReportState {
```
tmorgner: Get rid of that. There is no need to have this interface implemented here, as no one who expects and IAsyncReportListener makes use of the IAsyncReportState functionality.

tmorgner: More explanation: Your listener is created in the "PentahoAsyncExecutor" and only references properties from the task. So logically, there is no need to let the "PentahoAsyncExecutor" create it. If you move the creation of the listener into your "PentahoAsyncReportExecution", then you can eliminate the setListener(..) method and you can be sure about how your *implementation* computes the state (and how to get it). 

The only thing you need is a UniqueID for the key. Funny enough, your PentahoAsyncReportExecution class has such an ID, which it can share with your listener. Give your IAsyncReportExecution a getter for it, and you can use it in the key for the Executor.

prakapenka: JobManager uses IAsyncReportState as a read-only state and that is the main reason of implementing such interface. The another - having this interface may (or may not) help to have ability to serialize report status - taking a bit close look at JobManager. So:

- IAsyncReportState  is read only interface
- We serialize report state (see JobManager.getState()) to json (possibly for xml) according to THIS interface
- we don't mess up 'read only serialization' with 'IAsyncReportListener' methods that do expose listener as a listener

Do we still need to get rid of IAsyncReportState?

"PentahoAsyncReportExecution" does not know it's own id if it has no listener. InstanceId - is ID nested from BackgroundJobReportContentGenerator. Remember task is:

1. queued
2. working
3. finished

so this states will be:

1. BackgroundJobReportContentGenerator->PentahoAsyncExecutor
2. PentahoAsyncExecutor
3. PentahoAsyncExecutor -> JobManager (retuns content)

I want to track a gap how long a report execution action lives between creation of callable and actual execution. This could be tracked (I assume could) with audit records with same instance ID: BackgroundJobReportContentGenerator creates callable and makes audit record with it's id. Then - callable has an instance id of BackgroundJobReportContentGenerator - it make record that execuiotn actually started. Listener UUID is different. Listener ID is UUID used by JobManager to track execution status. So instanceId - is for audit purposes, listener id - to determine execuiton id which is different. I am not sure I can use ContentGenerator ID as ID for async job - so we have 2 separate ids now. Listener do not share instanceId with PentahoAsyncReportExecution.

Is it sounds reasonable?

prakapenka:
close this since it was submitted in another pull request.

PR #248, opened 2016-03-17
[BACKLOG-6878] RV: Reviewed init usages

annapopova1:

### DIFF
--- a/test-js/reportviewer-spec.js
+++ b/test-js/reportviewer-spec.js
@@ -40,8 +40,6 @@ define(["reportviewer/reportviewer", "reportviewer/reportviewer-logging", "repor
 
           var reportPrompt = new Prompt();
 
-          spyOn(reportPrompt, 'initPromptPanel').and.callFake(function () {});
-
           spyOn($, 'ajax').and.callFake(function (params) {
             params.success(parameterDefinition);
           });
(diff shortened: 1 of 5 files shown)


diogofscmariano:
Nice code clean up!

test-js/reportviewer-spec.js
```diff
@@ -12,170 +12,210 @@
...
-              expect(result).toBe(options);
-              expect(result['name']).toBe(options['action']);
-            });
-          });
+        // This test checks subscription on afterUpdate event.
+        it("should update layout after refreshing prompt panel", function(done) {
+          spyOn(reportViewer.view, "updateLayout");
+
+          reportViewer.reportPrompt.api.operation.refreshPrompt();
+          done();
```
jvelasques: @annapopova1 since the expect is after the done is called, it won't actually count towards the success/failure of the test.
This needs to be revisited, it seems to be causing one other test to fail when using karma.ci.conf.js

annapopova1: Thank you @jvelasques . I will check this test.

annapopova1: @jvelasques tests were fixed with https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/254

PR #250, opened 2016-03-17
[BACKLOG-7147] - Replaced paginate property reads with state calls. U…

diogofscmariano:
…pdated tests

[BACKLOG-7148] - Replaced totalPages property reads with state calls. Updated tests
[BACKLOG-7149] - Replaced page property reads with state calls. Updated tests


diogofscmariano:

PR #258, opened 2016-03-18
[BISERVER-10627] As an end user when a report run is in progress, I w…

Aliaksandr-Kastenka:
…ant to cancel it(Provide UI and implement JS logic)

Goes with https://github.com/pentaho/pentaho-commons-gwt-modules/pull/367
@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -42,6 +42,8 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
       get reportPrompt () {
         return _reportPrompt;
       },
+      _currentReportStatus: null,
+      _currentReportUuid: null,
 
       _bindPromptEvents: function() {
         var basePromptReady   = this.reportPrompt.ready.bind(this.reportPrompt);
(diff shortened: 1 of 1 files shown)


tmorgner:
You are referencing "window.viewer" all the time, which is not necessary. Reference your report-viewer's current instance via "this" and you will be equally fine and wont have to rely on the top-level window bound variables at all. It will also make your code clearer to read, as it is now clear that you are calling your own instance.

Your cancel-code is rather verbose. It also exists as the exact same code in 3 different locations. I think it is time to create a new function on the report-viewer called "cancel()" for that code. 

You are installing the cancel handler into the reportContent iframe once. But when a frame or window gets new content, all old code will be removed. 
Put the boolean flag that tracks whether you already installed the handler into the window object of the iframe. This way both variables travel together and you can detect lost handlers.

Aliaksandr-Kastenka:
> You are referencing "window.viewer" all the time, which is not necessary. Reference your report-viewer's current instance via "this" and you will be equally fine and wont have to rely on the top-level window bound variables at all. It will also make your code clearer to read, as it is now clear that you are calling your own instance.

I am using "this" where it is possible, in other cases - "window.viewer", for example in this scope:
 window.onbeforeunload = function(e) {...}
this._currentReportStatus is undefined, window.viewer._currentReportStatus is correct value.

@tmorgner 
Please take a look.

tmorgner:
This is undefined as event-handlers dont have a "this". (Its weird, but that's the way browsers work.) 
You can fix that easily by using the bind(this) method on your functions.

         window.myEvent = function(e) { .. }.bind(this);

Aliaksandr-Kastenka:
@tmorgner

PR #261, opened 2016-03-21
[BACKLOG-7169],  	[BACKLOG-7181], junit test clean up.

prakapenka:
This brings:

- tests cleanup - since we don't fork junits we have side effects on code from one test attempts to set somthing to a static context, while in other tests this static context unexpectedly becomes available.
- [BACKLOG-7181] - use mega-greed catch block to catch exception as far as errors. Errors like NoSuchMethod error while jars incompatible with dependencies. We should avoid hang on working status.
- [BACKLOG-7169] - fix job manager to use proper content headers.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/JobManager.java
+++ b/src/org/pentaho/reporting/platform/plugin/JobManager.java
@@ -25,6 +25,8 @@ import org.codehaus.jackson.map.ObjectMapper;
 import org.pentaho.platform.api.engine.IPentahoSession;
 import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
+import org.pentaho.platform.util.RepositoryPathEncoder;
+import org.pentaho.platform.util.web.MimeHelper;
 import org.pentaho.reporting.platform.plugin.async.IAsyncReportState;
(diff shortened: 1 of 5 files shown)


prakapenka:
@tmorgner  please review.

prakapenka:
@tmorgner tag!

src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -111,15 +114,46 @@ public JobManager( boolean isSuportAsync ) {
...
+    response = calculateContentDisposition( response, state );
+
+    return response.build();
+  }
+
+  private Response.ResponseBuilder calculateContentDisposition( Response.ResponseBuilder response,
+      IAsyncReportState state ) {
+    String extension = MimeHelper.getExtension( state.getMimeType() );
+    String path = state.getPath();
+    String fileName = path.substring( path.lastIndexOf( "/" ) + 1, path.lastIndexOf( "." ) );
```
tmorgner: Please make this code safe. Your path may nor have a "/" or "." character. If you want a safe option for that, use IOUtils#getFileName and IOUtils#stripFileExtension from libbase. It contains the necessary checks to handle the -1 index returned by "lastIndexOf" in case the search string is not found.

prakapenka:
@tmorgner  fixed according to comments, please review.

PR #263, opened 2016-03-21
[BACKLOG-7204] Remove total page update on each status and update 

EgorZhuk:
We don't need it because we will have a popup with status updates soon.
Also it is useless anyway - we update cache only when report is finished.

Thomas, please review.

@tmorgner


tmorgner:
I agree that updating the pages on _every_ status update is probably overkill. 

But:
This change would leave the page control in an invalid (or at least indeterminate) state between the first-page being displayed and the report finishing. 

I would say have at least one such update when you first encounter the "first-page" content. This way you get the proper update to make the UI consistent, without having to do the same update over and over again after that.

EgorZhuk:
Will implement "write cache on request" logic

PR #264, opened 2016-03-21
[BACKLOG-6764, BACKLOG-6885] - Removing unnecessary this.panel check

jvelasques:


jvelasques:

krivera-pentaho:
I know i had an issue retrieving the parameter definition when i did not have this check. are you sure you can remove it without ill effect?

jvelasques:
I'm pretty sure, we will have an [api](https://github.com/pentaho/pentaho-platform-plugin-reporting/blob/master/package-res/reportviewer/reportviewer-prompt.js#L21), which will have a [promptPanel](https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/api/OperationAPI.js#L29), the panel surely has a [dashboard](https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/package-res/resources/web/prompting/PromptPanel.js#L507). It makes perfect sense that we do not need the check. Making sense is not enough though, so I tested in a server, forcing the error before [_createPromptPanelFetchCallback](https://github.com/pentaho/pentaho-platform-plugin-reporting/blob/master/package-res/reportviewer/reportviewer-prompt.js#L128), had no issues :)

krivera-pentaho:
Excellent. Thank you!

PR #265, opened 2016-03-21
[BACKLOG-7199] Fatal error occurs in case async report generation and…

dzmitry-hryhoryeu:
… empty page number text box
@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -410,10 +410,12 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
           }.bind(this));
         },
 
-        pageChanged: function(pageNumber) {
-          this._setAcceptedPage('' + (pageNumber - 1));
-          this.reportPrompt.api.operation.submit();
-        },
+          pageChanged: function (pageNumber) {
+              if (!isNaN(pageNumber)) {
+                  this._setAcceptedPage('' + (pageNumber - 1));
+                  this.reportPrompt.api.operation.submit();
+              }
+          },
 
         togglePromptPanel: function() {
           this.showPromptPanel(registry.byId('toolbar-parameterToggle').checked);


lucboudreau:
http://10.100.11.222:8080/job/wingman-osgi/69/

lucboudreau:
Base build failed with return code 1

dzmitry-hryhoryeu:
You have incorrect pentaho-reporting-engine dependency

lucboudreau:
@dzmitry-hryhoryeu Not sure what you mean. If the project is not depending on the right version, please have it fixed.

tmorgner:
@lucboudreau This project requires a different build setup. You would need to set the "dependency.pentaho-reporting.revision" to 7.0-EWOK-SNAPSHOT to build it. Not sure if Wingman can do that. 

We are not working against master, and building for an parallel integration branch with their own artifact-ids. The merge into master is scheduled for somewhere next month.

tmorgner:
Dzimity, this actually does not prevent the system from entering an invalid state. You just worked around a symptom. 

The "pageChanged" method is a callback registered in the page-control component. That component calls this callback when someone calls "setPageNumber" on that control.

So you can prevent the error that happens deep inside the callback if you prevent "setPageNumber" from receiving invalid values. The page-control gets updated via the "updatePageControl" method, which gets the total-number-of-pages and the current-page from the prompt-panel. 

The prompt-panel API currently does not allow you to change that state easily. (Yes, we could hack and modifiy the private properties of the prompt panel, but lets fix it properly.)

Modify the "PromptPanel.js" method "setState" to allow updates of the "totalPages" property by adding the following line there. Also please update the documentation of that method to inform others why this state property can be changed:

> (state.totalPages != null) && this.paramDefn.totalPages = state.totalPages;

Then, in reportviewer.js in the _updateReportContentCore method, you can replace the 

        var pageContr = registry.byId('pageControl');
        pageContr.setPageCount(resultJson.page);

with 

      reportPrompt.api.operation.state({totalPages: 10});
      updatePageControl();

This then both updates the internal state of the prompt-panel and updates the page-control.
Add some sanity tests so that you only make the update if the json actually contains a valid page-total, and you are safe forever.

lucboudreau:
@tmorgner 10-4.

PR #266, opened 2016-03-22
[BACKLOG-7264] Added progress bar showing threshold configuration

EgorZhuk:
Set default to 1,5 seconds, otherwise dialog blinks.
Also switched to the default constructor in spring.

Thomas, please review.
@tmorgner


tmorgner:
Looks great. Merged.

PR #268, opened 2016-03-22
[BACKLOG-6961] Fixed usage of underscore dependency

afrjorge:
- using "common-ui/underscore" instead of "underscore" module ID


afrjorge:

PR #269, opened 2016-03-23
Backlog 7352 - Fix '-1' row quantity in the end of generation

EgorZhuk:
AbstractReportProcessor creates new event when calling listener.reportProcessingFinished/
This new event has default values for the fields, so we can ignore them and just update the status.

Thomas, take a look please.

@tmorgner

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
+++ b/src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
@@ -171,9 +171,11 @@ public class CachingPageableHTMLOutput extends PageableHTMLOutput {
         logger.warn( "Using cached report data for " + key );
         final ReportProgressListener listener = ReportListenerThreadHolder.getListener();
         if ( listener != null ) {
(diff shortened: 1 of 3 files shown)


src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -171,9 +171,11 @@ public synchronized int generate( final MasterReport report, final int acceptedP
         logger.warn( "Using cached report data for " + key );
         final ReportProgressListener listener = ReportListenerThreadHolder.getListener();
         if ( listener != null ) {
-          listener.reportProcessingFinished(
+          final ReportProgressEvent event =
             new ReportProgressEvent( "Cache", ReportProgressEvent.GENERATING_CONTENT, 0, 0,
```
tmorgner: Events in Java use the first parameter in their constructor to tell who is generating the event. So the string "Cache" is not the originator, your CachingHTMLReportOutput instance is. So please change that to "this".

EgorZhuk:
@tmorgner done

PR #270, opened 2016-03-23
[BACKLOG-7204] Get page from existing generation, don't start new one

EgorZhuk:
Thomas, could you check if I'm on the right way.
Current implementation doesn't store pages into cache on each user interaction (it really speeds up navigation) but it could be easily changed.
@tmorgner

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
+++ b/src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
@@ -17,9 +17,12 @@
 package org.pentaho.reporting.platform.plugin.output;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.pentaho.platform.api.engine.IApplicationContext;
+import org.pentaho.platform.api.engine.IPentahoSession;
+import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 import org.pentaho.reporting.engine.classic.core.MasterReport;
 import org.pentaho.reporting.engine.classic.core.ReportProcessingException;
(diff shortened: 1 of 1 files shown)


tmorgner:
Storing multiple listeners wont work, as the SimpleReportingComponent produces new instances of the CachingPageableHTMLOutput all the time. And there is no need for it anyway, as we can access the job instance while it is running via the AsyncExecutor. 

I think you can get to it easier with this approach:
 
First to know: If you produce page 10, then implicitly that means pages 1 to 9 are available too. So if we update the "accepted-page" variable to a larger number and just populate the cache one more time, we implicitly have cached content for all previous pages too.

The report-viewer today simply listens for the state change from "WORKING" to "CONTENT_AVAILABLE". It is a boolean condition and does not indicate what page is available in the cache. So we need to expand the job-status a bit, for instance by adding a "generatedPage" property. This would then reflect what page is actually available in the cache (vs the plain "page" property, which tells us what the reporting engine has generated (but which may not yet be in the cache)).

We then not just check for the plain CONTENT_AVAILABLE in the reportviewer, but also check whether the selected page from the pagination-control is within the "generatedPage" range before we fire the request to fetch the content. 


Then we can implement a simple loop:

While the report is not fully processed:
(1) User requests page.
(2) Check if page is available as accessible content ( status.generatedPage >= requestedPage )
   (a) yes: Directly request the page. The request will succeed quickly as the content is available in the cache. 
   (b) no: Update the existing job on the server to produce the page. Then show the progressbar again.
(3) periodically request status updates and wait for ( status.generatedPage >= requestedPage ) to become true. Then request that page.

While the report-viewer is processing the request, it holds a reference to the job via the job-id. So we can update the existing job without any problem. Lets say that id is "123". 

If the user navigates to a new page (lets say 10) while the job runs in the background, fire a request to the job-manager and request it to update the AsyncReportExecution with the new requested page. 

/jobs/123/requestPage?page=10

Internally, we can then update the AsyncListener to the new requested page. 

    PentahoAsyncExecutor.requestPage(UUID("123"), 10);

which then calls 
      tasks.get(UUID("123")).requestPage(10);

which then updates the AsyncReportStatusListener with the new accepted page. That needs a new property on it ("requestedPage"), but nothing awfully bad here.

Within the CachingHTMLReportOutput, instead of using just the "accept-page" parameter, use Math.Max( acceptedPage, asyncReportListener.getRequestedPage()) as your condition to decide when to write content to the cache in "CachingPageableHTMLOutput.java:103".

EgorZhuk:
> Storing multiple listeners wont work

It's static map :smile: 
Thank you Thomas, I will follow proposed approach.
@tmorgner

PR #272, opened 2016-03-24
[BACKLOG-7404] Async mode still fires expensive pagination call

Aliaksandr-Kastenka:
@tmorgner 
Please review.


tmorgner:
Thank you!

PR #273, opened 2016-03-24
[BACKLOG-7410] - Fatal Error occurs on attempt to press Cancel button in progress bar

prakapenka:
- do not erase report 'CANCELED' status in case of Exception

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
+++ b/src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
@@ -107,7 +107,10 @@ public class PentahoAsyncReportExecution implements IAsyncReportExecution<InputS
   }
 
   private void fail() {
-    listener.setStatus( AsyncExecutionStatus.FAILED );
+    // do not erase canceled status
+    if ( !listener.getState().getStatus().equals( AsyncExecutionStatus.CANCELED ) ) {
+      listener.setStatus( AsyncExecutionStatus.FAILED );
+    }
     AuditHelper.audit( safeSession.getId(), safeSession.getId(), url, getClass().getName(), getClass().getName(),
         MessageTypes.FAILED, auditId, "", 0, null );
   }


prakapenka:
@tmorgner please review. Tag!

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
```diff
@@ -18,13 +19,13 @@
...
-  private final SimpleReportingComponent reportComponent;
-  private final AsyncJobFileStagingHandler handler;
-  private final String url;
-  private final String auditId;
-  private final IPentahoSession safeSession;
+  protected final SimpleReportingComponent reportComponent;
+  protected final AsyncJobFileStagingHandler handler;
+  protected final String url;
+  protected final String auditId;
+  protected final IPentahoSession safeSession;
```
tmorgner: There is no need to set these variables as protected. Good encapsulation rules say that all data should be as invisible as possible. In the scope of this bug fix, there is no reason for this change.

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
```diff
@@ -28,7 +28,7 @@
  * Simple synchronized bean with async execution status. This class is part of the PentahoAsyncReportExecution
  * implementation and should not be used outside of that.
  */
-class AsyncReportStatusListener implements IAsyncReportListener {
+public class AsyncReportStatusListener implements IAsyncReportListener {
```
tmorgner: The AsyncReportStatusListener is an implementation details of the AsyncReportExecution and is not useful outside of that context. There is no need to make that class public.

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
prakapenka: There is a need to set these variables as protected since they are going to be re-used in PIR plugin. As far as making listener public. Since PIR don't follow reporting package structure, but has it as dependency.

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
prakapenka: As far as we a going to have async mode for PIR....

prakapenka:
@tmorgner please review, thank you in advance.

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
tmorgner: If you are reusing them there, make that in a separate commit related to that case. Dont wildly mix work on two cases - if you work on two cases at the same time, use separate branches to keep that work cleanly separate.

prakapenka:
@tmorgner please review

PR #274, opened 2016-03-24
[CLEAN-UP] - make classes pulic, members protected to reuse code in PIR.

prakapenka:
...remove some restrictions to have ability to reuse classes in different package.


prakapenka:
@tmorgner please review. Tag!

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
```diff
@@ -18,13 +19,13 @@
...
-  private final String auditId;
-  private final IPentahoSession safeSession;
+  protected final SimpleReportingComponent reportComponent;
+  protected final AsyncJobFileStagingHandler handler;
+  protected final String url;
+  protected final String auditId;
+  protected final IPentahoSession safeSession;
 
-  private AsyncReportStatusListener listener;
+  protected AsyncReportStatusListener listener;
```
tmorgner: I'm fine with "protected final" members being accessed directly, as there is no harm anyone can do to the internal state of the class. But for the "listener" please use getters/setters to control access. If you only need read access to the listener, then make sure you only provide a getter.

Also: I'd like to see the context of these changes so I'll hold off merging this change until there is some PIR code showing how these protected members are used. As you are working in separate branches for each feature you work on, that should not block you from developing the features.

prakapenka:
@tmorgner find a way to avoid increasing visibility of listener.
But still a lot of new clean-ups.
Please review.
Tag!

prakapenka:
@tmorgner please review.

src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -18,40 +18,28 @@
...
 import org.pentaho.reporting.platform.plugin.async.PentahoAsyncExecutor;
 
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.core.CacheControl;
+import javax.ws.rs.*;
```
tmorgner: [Pentaho Coding standards](http://wiki.pentaho.com/display/PEOpen/Code+Formatting+and+CheckStyle) say: No star imports. You can get the template for IntelliJ from [here](https://github.com/pentaho/pentaho-coding-standards)

tmorgner:
There is a spelling error in the class name of "AbstractJobManager". The object-mapper you expose as protected member variable should be private with a getter. I do not think protected variables are sane, as you cannot control what happens to them.  And I would say to make it safe in all environments (ie when a OEM tries to subclass this jobmanager), make it a ObjectWriter instead of a ObjectMapper (ie new ObjectMapper().writer()). *Then* you can go even further and make it "protected static final" as ObjectWriter is guaranteed to be thread-safe. 


IAsyncReportExecution#getBytes is not used anywhere.

IAsyncReportState does and should not be the place to expose the bytes. The bytes-counter is a derived property of the produced stream, which is a result of the ReportExecution, not the state-monitoring. 

It would be cleaner to return the input-stream and the bytes in one operation if you need both of them. As we have no simple Tuple type in Java, we have to add a custom value object for that.

For that, create an interface for that:

public interface IFixedSizeStreamingContent {
   public InputStream getStream();
   public long getContentSize();
}

and implement that interface either as a value-object or implement it on AsyncJobFileStagingHandler, which contains both the stream and the size. Finally make your IAsyncReportExecutor return futures of IFixedSizeStreamingContent instead of the plain InputStream. 

Then you can get rid of the AsyncReportCountBytesWrapper and the changes to the IAsyncReportState.


AbstractPentahoAsyncExcutor: All methods dealing with the CompositeKey should be protected, even the static ones. There is no need to allow outsiders to call these methods. PentahoAsyncExecutor adds itself as LogoutListener, but this is already done by the AbstractPentahoAsyncExecutor. Remove the duplicate call. 

I also doubt the need for IAsyncExecutor interface. None of the methods of this interface are ever called upon, and I cannot see how the methods in there can be sensibly separated from all other methods in IPentahoAsyncExecutor. Make AbstractPentahoAsyncExcutor implement the IPentahoAsyncExecutor interface and leave the methods you cannot implement here as abstract.

prakapenka:
All methods dealing with the CompositeKey should be protected, even the static ones.
->
protected static 
->
http://stackoverflow.com/questions/24289070/why-we-should-not-use-protected-static-in-java

prakapenka:
" Make AbstractPentahoAsyncExcutor implement the IPentahoAsyncExecutor interface and leave the methods you cannot implement here as abstract." - The reason to create an abstraction was the fact that for PIR we can't (I did not found a way) just to reuse IAsyncReportExecution<T> interface. We need to extend it - so we need also to extend executor class as far as jobManager class. While we comfortable with Future<InputStream>. Just to make attempt to eliminate duplication I made attempt to move all future-related code up to abstract class and leave all 'IAsyncReportExecution<T>' in concrete implementation.

tmorgner:
If you dont want protected static final fields, then use a *private* static final field and write a *protected* static getter - it has the same effect. But there is no need to make these methods public - they are an implementation detail and therefore should not be exposed to ordinary callers of the code.

tmorgner:
> " Make AbstractPentahoAsyncExcutor implement the IPentahoAsyncExecutor interface and leave the methods you cannot implement here as abstract." - The reason to create an abstraction was the fact that for PIR we can't (I did not found a way) just to reuse IAsyncReportExecution interface. We need to extend it - so we need also to extend executor class as far as jobManager class. While we comfortable with Future. Just to make attempt to eliminate duplication I made attempt to move all future-related code up to abstract class and leave all 'IAsyncReportExecution' in concrete implementation.

If you need to extend it, means you add new functionality, but dont remove functionality in a way that breaks polymorphism, then simply use a sub-interface and generics to avoid the casting. However, to give a definite answer I would  need to see the code in PIR where you would want to use it.

prakapenka:
.. really, I don't know why I did not make it with generics at first time. Exactly generics is recipe.

tmorgner:
Please squash these 12 commits so that any obsolete and now corrected changes are no longer there. 

There are multiple commits dealing with the JobManager, for instance, with old commits that are not correct and that have been overwritten in a later commit. Squash all into one or reorder and squash the commits along logical boundaries.

prakapenka:
squashed and hang on waiting for approve.
@tmorgner  please check https://github.com/prakapenka/pentaho-platform-plugin-interactive-reporting/commits/PIR-1225 for changes for PIR.

Thank you in advance.

src/org/pentaho/reporting/platform/plugin/AbstarctJobManager.java
```diff
@@ -0,0 +1,168 @@
...
+import javax.ws.rs.core.StreamingOutput;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * Created by dima.prokopenko@gmail.com on 3/30/2016.
+ */
+public abstract class AbstarctJobManager {
+
```
tmorgner: This class still has a spelling error here. There is no such word as "Abstarct". It is "Abstract".

prakapenka: Fixed!

prakapenka:
update with excellent fixes from @tmorgner + check-style from my side.
@tmorgner please review.

PR #275, opened 2016-03-25
Backlog 7204

EgorZhuk:
This one is about navigation through still running report.

How it works on UI:

1. ...
2.  User requests a new page.
3.  If generation have already been finished or page is in cache - just fire a job to get the page from cache. :ok_hand: 

   If not - send a request to the generation task to store the cache.
   When you see that page is now in cache - just fire a job to get it.

How it works on server side:

1. ...
2.  Somebody requested a page.
3.  If page is already available in processor - write to cache as much pages as you have and update generatedPages value.

   If not listen to events and do the job when it is.

I've recorderd a short [demo](https://dl.dropboxusercontent.com/u/71148577/page_navigation.webm)

Thomas, please take a look.
@tmorgner

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -44,6 +44,8 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
       },
       _currentReportStatus: null,
       _currentReportUuid: null,
+      _currentStoredPagesCount: null,
+      _requestedPage: -1,
 
       _bindPromptEvents: function() {
         var basePromptReady   = this.reportPrompt.ready.bind(this.reportPrompt);
@@ -386,6 +388,15 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
          */
         _setAcceptedPage: function(pageNumber) {
           this.reportPrompt.api.operation.setParameterValue("accepted-page", pageNumber);
+        },
+
+        /**
+         *
+         * @param pageNumber
+         * @private
+         */
+        _getAcceptedPage: function() {
+          return this.reportPrompt.api.operation.getParameterValues()["accepted-page"];
         },
 
         /**
@@ -856,6 +867,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
                 dlg.setText3(_Messages.getString('FeedbackScreenRow') + ': ' + resultJson.row + ' / ' + resultJson.totalRows);
                 this._currentReportStatus = resultJson.status;
                 this._currentReportUuid = resultJson.uuid;
+                this._currentStoredPagesCount = resultJson.generatedPage;
 
                 progressBar.set({value: resultJson.progress});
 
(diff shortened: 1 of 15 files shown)


src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncExecutor.java
```diff
@@ -108,16 +103,28 @@ private String getSessionId() {
...
+
   @Override public IAsyncReportState getReportState( final UUID id, final IPentahoSession session ) {
+    validateParams( id, session );
+    // link to running task
+    final IAsyncReportExecution<InputStream> runningTask = tasks.get( new CompositeKey( session, id ) );
+    return runningTask == null ? null : runningTask.getState();
+  }
+
+  private void validateParams( final UUID id, final IPentahoSession session ) {
     if ( id == null ) {
```
tmorgner: You can safe some lines by using ArgumentNullException.validate(..) which also gives you a nice standard message.

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -97,10 +97,21 @@ public CacheListener( final String key, final int acceptedPage,
...
+      if ( asyncReportListener.getRequestedPage() > 0 ) {
+        //user requested new page before original generation finished
+        //Generation could already pass the page but we need to update cache anyway
+        page = Math.max( asyncReportListener.getRequestedPage() + 2, reportProgressEvent.getPage() );
+      } else {
+        //it's just an original request
+        page = acceptedPage + 2;
+      }
+
+      final boolean requestedPageIsStored = reportProgressEvent.getPage() == page;
```
tmorgner: You can get multiple progress events per page. It might be wise to have a flag that remembers the last page you persisted. This way you only persist a given page-set once. 

Adding a new member variable here for "lastAcceptedPageWritten" and checking against that as well as your accepted-page will save you from storing the same content multiple times.

package-res/reportviewer/reportviewer.js
```diff
@@ -863,30 +875,44 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
-                        $('#reportContent').attr("data-src", urlContent2);
-                        this._updatedIFrameSrc = true;
-                        dlg.hide();
-                        isIframeContentSet = true;
-                      }
-                    });
-                    //get first page
-                    pentahoGet('reportjob', url.substring(url.lastIndexOf("/report?")+"/report?".length, url.length), handleContAvailCallback, 'text/text');
+
+                    if(this._currentStoredPagesCount > me.view._getAcceptedPage()){
```
tmorgner: Why do you use me.view._getAcceptedPage here and this.requestedPage later. Shouldnt that be consistent, for instance by ensuring that you always use "this._requestedPage"?

package-res/reportviewer/reportviewer.js
```diff
@@ -863,30 +875,44 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
+                      pentahoGet('reportjob', url.substring(url.lastIndexOf("/report?")+"/report?".length, url.length), handleContAvailCallback, 'text/text');
+                    }
+
                     var urlStatus = url.substring(0, url.indexOf("/api/repos")) + '/plugin/reporting/api/jobs/' + resultJson.uuid + '/status';
                     setTimeout(function(){ pentahoGet(urlStatus, "", handleResultCallback); }, this.reportPrompt._pollingInterval);
                   } else {
+
+                    if( (this._requestedPage > 0) && (this._currentStoredPagesCount > this._requestedPage) ){
```
tmorgner: requestedPage has an initial value of "-1" to signal no selection. Then here you test for > 0, which means there is a possible value "0" that is not covered. Assuming the page-control cannot have 0 as valid value, why not use that instead of -1.

package-res/reportviewer/reportviewer.js
tmorgner: You may have to set "this.requestPage" to a valid state based on the user's selection in the page-control when you first start the report-content fetching to make it consistent, but I feel it would add to the clarity of the code-flow.

EgorZhuk:
Fixed.
@tmorgner

EgorZhuk:
Thank you, Thomas.

PR #284, opened 2016-03-29
[BACKLOG-7481] Clean leaked future

EgorZhuk:
I've realised that we don't clear futures for main report generation job.
Added a clenup callback.

Thomas, please review.
@tmorgner

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -935,6 +935,9 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
                     this._updatedIFrameSrc = true;
 
                     dlg.hide();
+                  } else {
+                    var urlClean = url.substring(0, url.indexOf("/api/repos")) + '/plugin/reporting/api/jobs/' + this._currentReportUuid  + '/clean';
+                    pentahoGet( urlClean, "");
                   }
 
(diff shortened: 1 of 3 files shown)


prakapenka:
This will clear the memory but will not clean staging content. To clean staging content from disk client must call .close() method on staging input stream. This means client should call

InputStream input = future.get();
and then immidiately 
input.close() - that will close input stream as far as will clean staging content from temp folder.

prakapenka:
IOUtils.closeQuietly( input )

tmorgner:
I would be careful with that wish. The close call should come naturally already when the client fetches the content in JobManager/content. 

If you close the stream here, you may interfere with that request, which will run in parallel as the IFrame loades. So either we delay  the "/cleanFutures" until the content has been fully fetched (which is hard to detect as IFrames dont give up that information easily and in case a plugin handles the content, we would not get that information at all), or we simply ensure that all content is cleaned up when we close the session (@see PentahoAsyncExecutor#onLogout).

EgorZhuk:
With first page feature switched on main job plays serving role - it generates report, updates statuses but we never request content from it. We just leave it waiting forever :cry: 
The same for canceed job - we don't perform cleanup.

I fixed it in #287 - note we call clean only for canceled jobs or jobs where iframe is already populated ( by the 1st page ).
@tmorgner

PR #287, opened 2016-03-30
[BACKLOG-7481] - Advanced cleanup

EgorZhuk:
More cleanup:

1) Clean future for canceled report 
2) Clean tmp files in cleanup call
3) Unused  isCanceled flag

Thomas, please take a look.
@tmorgner

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -806,7 +806,6 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         this._isReportHtmlPagebleOutputFormat=me.view._isHtmlPagebleOutputFormat(options['output-target']);
 
         //BISERVER-1225
-        var isCanceled = false;
         var isFirstContAvStatus = true;
         var isIframeContentSet = false;
         var isFinished = false;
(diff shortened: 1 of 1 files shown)


tmorgner:
Agree with the cancel fireing the clean request. 

But looking at the big picture on the "FINISHED" case, I sense a race condition.

This clean up could introduce a race condition between the iframe and the clean-up. After all, we cannot control the order in which the browser sends the requests or in which order the server handles them. So it is possible that your clean gets executed before the "/content" request. 

Wouldn't it be better and safer to keep relying on the "job/../content" handler to close and clean up the staged content?

EgorZhuk:
@tmorgner  now I see, totally agree you. Let's rely on "job/../content"  and org.pentaho.reporting.platform.plugin.async.PentahoAsyncExecutor#onLogout

EgorZhuk:
@tmorgner only js changes, no is.close() in JobManager

package-res/reportviewer/reportviewer.js
```diff
@@ -951,16 +954,20 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
 
                     dlg.hide();
                   } else {
-                    var urlClean = url.substring(0, url.indexOf("/api/repos")) + '/plugin/reporting/api/jobs/' + this._currentReportUuid  + '/clean';
```
tmorgner: Can you remove that "/clean" call. It will trigger the race otherwise, as /clean removes the future, and the /content call can then not find the content on the server.

EgorZhuk:
Then it's just code cleanup. Will inclued it in another PR.

PR #289, opened 2016-03-30
[BACKLOG-7527] - Prevent updating content twice 

EgorZhuk:
Hello Thomas,

I've played with navigation for a while and noticed some redundant calls.
Here are the fixes:

1) If page is available and process is running we don't need to manually spawn another job it will be spawned by the main job ([line](https://github.com/EgorZhuk/pentaho-platform-plugin-reporting/blob/511f10ada2a9f9d24fe7da206538c1d959263c00/package-res/reportviewer/reportviewer.js#L922)).

2) Add another check to special case block ([line](https://github.com/EgorZhuk/pentaho-platform-plugin-reporting/blob/511f10ada2a9f9d24fe7da206538c1d959263c00/package-res/reportviewer/reportviewer.js#L958)) to avoid updating content twice.

@tmorgner


EgorZhuk:
@tmorgner  added cancelation of the running job if parameters/output target changed

PR #292, opened 2016-03-31
[BACKLOG-7180] The wait dialog “Just a few moments please” appears be…

dzmitry-hryhoryeu:
The wait dialog “Just a few moments please” appears before the progress bar

Thomas, please review.
@tmorgner


tmorgner:
Looks good. Merged.

PR #294, opened 2016-03-31
[BACKLOG-6945] RV: Review refresh usages

annapopova1:
Added changes to remove last usages of panel object. So need merge after https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/721


package-res/reportviewer/reportviewer.js
```diff
@@ -639,35 +637,19 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
-        // this call actually gets executed **before** the logic fired by the focusout path,
-        // and fetchParameterDefinition has not been called yet.
-        this.reportPrompt.clicking = true;
+      beforeUpdateCallback: function() {
+        this.reportPrompt._isUpdatingPrompting = true;
+      },
+
+      afterUpdateCallback: function() {
+        this.view.updateLayout();
+        this._updateReportContent();
```
krivera-pentaho: do you need to add "this.reportPrompt._isUpdatingPrompting = false;" in this call?

annapopova1: No, we don't. We need remove this property after all callbacks and as a result I do it in the _submitReportEnded function.

package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -284,13 +247,6 @@ define([
           me.hideGlassPane();
         }
 
-        if (me.clicking) {
-          // If "Upgrading" a Change to a Submit we do not want to process the next Submit Click, if any
-          me.ignoreNextClickSubmit = (promptMode === 'USERINPUT');
```
diogofscmariano: Not sure if we cant remove this....

annapopova1: If I understand correctly this logic needs to avoid extra requests when we click on submit button during processing. But now I do not see a reason to save this logic because I do not find a way to click on this button during processing (glass pane exists during processing)...

diogofscmariano: We need to be very careful with this. we have many problems with this in the past i just want to make sure we dont face those issues again. On the same level, we have upgraded the lifecycle, and many of those issues might be over. Lets try it

package-res/reportviewer/reportviewer.js
```diff
@@ -639,35 +637,19 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
-        // we upgrade that request to behave like if a button had been clicked, by setting prompt.mode to 'MANUAL'.
-        // This way, if the user takes longer to release the mouse button,
-        // than the focusout response takes to replace the being clicked button by another one (Dashboards.init recreates everything),
-        // causing the click event no be generated,
-        // the focusout response processing will behave as if it were the response of a click.
-        //
-        // Because Dashboards.processChange ends up calling postChange in a setTimeout...
-        // this call actually gets executed **before** the logic fired by the focusout path,
-        // and fetchParameterDefinition has not been called yet.
-        this.reportPrompt.clicking = true;
```
diogofscmariano: We need to be very careful with this....

PR #299, opened 2016-04-05
[BACKLOG-6740] - Run in background feature

EgorZhuk:
Hello Thomas,

I've used Guava ListenebleFuture and Futures.addCallback to implement scheduling mechanism.
Unfortunately MoreExecutors.listeningDecorator ignores custom logic, so I've implemented my own decorator that allows us to extend ListenableFuture with desired logic (cancel support in our case). I've tried to make it easy to add more extensions in future.

Also I've made AsyncExecutionStatus more deterministic.

@tmorgner


tmorgner:
Home directory: 
No need to hard-code anything here. Give OEMs a chance to apply logic to that (ie: CEO's reports go into home directory, normal user's reports go elsewhere etc).

Create a "ISchedulingDirectoryStrategy" interface and inject it via the spring.xml. This way an OEM can happily provide their own implementation as needed. 

And as for your calculation of the home directory: There is code for that. Use the static methods of class "ClientRepositoryPaths" for accomplishing that same task. OEMs can provide their own implementation here, computing it yourself can therefore break that scenario.

EgorZhuk:
@tmorgner please take a look

tmorgner:
Scheduling dir property and constructor parameter is no longer used. Can be removed from PentahoAsyncExecutor. 

In WriteToJcrTask#call, the extension returned by MimeHelper.getExtension can be null. You would produce "contentnull" in the worst case scenario. 

Other than that it looks great.

EgorZhuk:
Hello Thomas,

I fixed review issues.
But actually there were more issues:

1) We didn't write all html pages to jcr in case pages were already in cache.
This situation can be reproduced by reducing poll capacity ( or running a lot of reports ).
You already have your report in cache but the content can't be returned immediately because all workers are busy.
You see progress bar and clicks "Run in background".

2) We need to prevent future cleanup, cancelation and directory cleanup if there are scheduled jobs for the session, otherwise we will have empty InputStream in callback.

3) I forgot to close InputStream for scheduled jobs, now fixed. Also I've added code that tries to clean up session staging dir if it is emty. The last job will remove it, otherwise we will have a lot of empty session directories.

4) Calling "/cleanup" from web without deleting staging file was a bad idea. 
We remove future from map and then won't be able to properly clean up staging dir.

5) We didn't close staging handler if report fails or is canceled. If we log out and there are job's in progress thebn files won't be closed and directory won't be removed.

6) Tech debt: shutdown method fot AsyncExecuror isn't executed, we will have 1 page in scheduled html if cache is disabled.

@tmorgner

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -221,6 +219,18 @@ public synchronized int generate( final MasterReport report, final int acceptedP
     }
   }
 
+  private int writeAllForScheduled( OutputStream outputStream, IReportContent cache ) throws IOException {
```
tmorgner: That will produce invalid content. We are producing full HTML files, so you append multiple full HTML documents together. That is not correct. If you'd want to merge HTML documents, you'd have to do it on an XML level (ie take the contents of the body-element and put them into the merged body. But then you'd probable also have to merge the CSS. That is a totally new world of trouble.

Additionally, imagine a merged document that used 1000 pages as source. That would be so big that most browsers would struggle to display them. 

I would say keep it as separate pages as before and then we talk with Kurtis on how this part of the feature should work. Somehow I have a feeling that run-in-background for pageable HTML is not a valid option.

EgorZhuk: @tmorgner I totally agree with you. But this is the way platform scheduler works for paginated html - it actually generates single page report. The difference is that in platform they choose output format before running the report, in our case the report is already running and we can't switch it to single page mode. I can't imagine the user that wants only one page of the report to be scheduled :)

EgorZhuk: @tmorgner removed it from commit

EgorZhuk:
@tmorgner added code to hide the button for paginated html

tmorgner:
Looking at it in depth there is one more thing: 

We can eliminiate the IDelegatedListenableFuture interface. Use ListenableFuture<T> directly. The extra interface does not add anything, as ListeneableFuture<T> is a Future<T> too.

PR #302, opened 2016-04-07
[BACKLOG-7597] - Allow to wait for content, optimize waiting time

EgorZhuk:
Corner case when staus is 'FINISHED' but the future is not done yet.
Don't see an issue in waiting for future completion on content request.
Also minimized amount of time to wait for this case.

@tmorgner  please review

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/JobManager.java
+++ b/src/org/pentaho/reporting/platform/plugin/JobManager.java
@@ -27,6 +27,7 @@ import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 import org.pentaho.platform.util.RepositoryPathEncoder;
 import org.pentaho.platform.util.web.MimeHelper;
+import org.pentaho.reporting.platform.plugin.async.AsyncExecutionStatus;
 import org.pentaho.reporting.platform.plugin.async.IAsyncReportState;
 import org.pentaho.reporting.platform.plugin.async.IPentahoAsyncExecutor;
 import org.pentaho.reporting.platform.plugin.async.PentahoAsyncExecutor;
@@ -97,14 +98,14 @@ public class JobManager {
     final IPentahoSession session = PentahoSessionHolder.getSession();
 
     Future<InputStream> future = executor.getFuture( uuid, session );
-    if ( future == null || !future.isDone() ) {
-      logger.warn( "Attempt to get content while execution is not done. Called by: " + session.getName() );
-      return get404();
-    }
 
     IAsyncReportState state = executor.getReportState( uuid, session );
     if ( state == null ) {
       return get404();
+    }
+
+    if( !AsyncExecutionStatus.FINISHED.equals( state.getStatus() ) ){
+      return Response.status( Response.Status.ACCEPTED ).build();
     }
 
     InputStream input = null;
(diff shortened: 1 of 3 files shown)


src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -97,10 +97,6 @@ public JobManager( final boolean isSupportAsync, final long pollingIntervalMilli
     final IPentahoSession session = PentahoSessionHolder.getSession();
 
     Future<InputStream> future = executor.getFuture( uuid, session );
-    if ( future == null || !future.isDone() ) {
```
prakapenka: Not checking future.isDone() we can potentially came into situation when we do request content future.get() and waiting for the competition. That will be a blocking operation. If we going to request content for 1st page calling /content end-point for content that is not ready it will be blocking request as it would be without async requests for first page. We may consider to request 'first page' in regular instead of async mode, asuming since 1st page is in cache - it is most probably will be available almost ready and we don't want to waste thread pool with such short-running tasks.

src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecution.java
```diff
@@ -82,9 +82,12 @@ public void notifyTaskQueued( final UUID id ) {
         AuditHelper.audit( safeSession.getId(), safeSession.getId(), url, getClass().getName(), getClass().getName(),
             MessageTypes.INSTANCE_END, auditId, "", ( (float) ( end - start ) / 1000 ), null );
 
+
+        final InputStream stagingContent = handler.getStagingContent();
+
         listener.setStatus( AsyncExecutionStatus.FINISHED );
 
-        return handler.getStagingContent();
+        return stagingContent;
```
prakapenka: Even in such way will not completely remove race condition since listener.setStatus(...) is synchronized+volatile - but sycnronized according to a listener class while future completition is in sync with future itself, not according to a listener.

tmorgner:
Changing the way the "FINISHED" state is set will reduce the window of the potential race condition, and that is a good one. 

At the JobManager change, I aggree that the new fully blocking way can have its pitfalls. A more staged approach may be in order here:

(1) If you get no future or state, return the 404. The job does not exist.

(2) If you get a state, check whether it is in the finished state. 

If it is NOT finished, you have an impatient client. Lets tell them to try again later. Send a 202 response.

> 202 Accepted
    The request has been accepted for processing, but the processing has not been completed. The request might or might not be eventually acted upon, and may be disallowed when processing occurs.

If the state is FINISHED, then - with the new changes here - the blocking on Future#get will be short. I am fine with just letting the caller wait until the Future has cleaned up. There are no expensive operations between the setting of the FINISHED state and the future having a result any more. 

The guaranteed calling sequence therefore is now:

(1) listener.setStatus(FINISHED)
(2) client requests state, sees finished result.
(3) client fetches page
(4) client potentially blocks on future (but that is OK, and I doubt it will ever happen thanks to network latency)
(5) Client gets content. Everyone is happy.

This new protocol ensures that a client that issued a /content request *after* the FINISHED status is set, it will always get the correct content. For any request issued before the FINISHED status is set, the result is pretty much undefined and relies on the caller handling the 202 correctly.

EgorZhuk:
@tmorgner  done

PR #303, opened 2016-04-07
[BACKLOG-7307] Progress Bar shows 'We are generating your report' sta…

Aliaksandr-Kastenka:
…tus in case report generation failed(error message added)

Error message added to server response in case when report generation is failed. It is shown inside of reportContent element, similarly as in case when async mode is disabled.

@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -934,9 +934,18 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
                     registry.byId('reportGlassPane').hide();
                   }
                 } else if (resultJson.status == "FAILED") {
-                  this.reportPrompt.showMessageBox(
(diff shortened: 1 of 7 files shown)


tmorgner:
First off: String concatenation of raw strings you get from the server? Bad idea!. If you want to display content from the server, you have to ensure that your content is properly encoded or you create XSS vectors. 

Use "'cdf/dashboard/Utils" to get access to the "Utils.escapeHtml(value)" function, which you can then use to create a safe message-string that can be used in the HTML document.

I also think that instead of using a popup-dialog if you have no message, use a default message instead and display it exactly like you would display a message that comes from the server. This way your UI will behave consistent in both cases.

However: I feel that our UX designers may not believe that is a sane idea to show raw error messages to the user in the first place. So please send a message to UX to ask for their opinion on this case and how we should display error information correctly.

PR #309, opened 2016-04-11
[BACKLOG-7552] Report generation is not cancelled by closing dashboard

Aliaksandr-Kastenka:
@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -604,10 +604,20 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
       }), // end view
 
       onTabCloseEvent: function (event) {
-        if (window && event.eventSubType == 'tabClosing' && event.stringParam == window.frameElement.id) {
-          this.cancel(this._currentReportStatus, this._currentReportUuid);
-          if(top.mantle_removeHandler) {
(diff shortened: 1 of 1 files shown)


package-res/reportviewer/reportviewer.js
```diff
@@ -604,10 +604,20 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
+        if(window.frameElement.src != null && window.frameElement.src.indexOf('dashboard-mode') !== -1 ){
+          if (event.eventSubType == 'tabClosing' && event.stringParam == window.parent.frameElement.id) {
+            this.cancel(this._currentReportStatus, this._currentReportUuid);
+            if(top.mantle_removeHandler) {
+              top.mantle_removeHandler(this._handlerRegistration);
+            }
+          }
+        }
+        else {
+          if (window && event.eventSubType == 'tabClosing' && event.stringParam == window.frameElement.id) {
```
tmorgner: "window" cannot be null. At least, if it is null, then your browser is so badly broken that your code will misbehave anyway.

If you expect window to be null, then your initial "if (window.frameElement ..)" test above would already fail with a NPE.

Aliaksandr-Kastenka:
@tmorgner 
Pull-request updated.

PR #314, opened 2016-04-13
[BACKLOG-7291] - Fix pagination control on dashboard

EgorZhuk:
Moved the logic of pagination HTML output detection right after the parameter definition parsing.
With latest master changes flag was not initialised properly.

@tmorgner please review


tmorgner:
.. and conceptually, that move also makes sense. We are working against the prompt-data, after all.

PR #315, opened 2016-04-13
[CLEANUP] - Fix test failing sometimes

EgorZhuk:
- we don't have control of execution order, let's make it more deterministic that write starts after read

@tmorgner please take look

### DIFF
--- a/test-src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackendTest.java
+++ b/test-src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackendTest.java
@@ -18,6 +18,7 @@
 package org.pentaho.reporting.platform.plugin.cache;
 
 import junit.framework.Assert;
+import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
@@ -45,9 +46,18 @@ public class FileSystemCacheBackendTest {
 
   private static FileSystemCacheBackend fileSystemCacheBackend;
 
+  @After
+  public void clearLatchesAfterEachTest() {
+    readStartedSignal = null;
+    writeStartedSignal = null;
+  }
+
+  private static CountDownLatch readStartedSignal;
+  private static CountDownLatch writeStartedSignal;
+
   @BeforeClass
   public static void setUp() {
-    PentahoSessionHolder.setSession( new StandaloneSession(  ) );
+    PentahoSessionHolder.setSession( new StandaloneSession() );
     fileSystemCacheBackend = new FileSystemCacheBackend();
     fileSystemCacheBackend.setCachePath( "/test-cache/" );
   }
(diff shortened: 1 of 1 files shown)


tmorgner:
No, a sleep is not really deterministic - you just decreased the likelihood of failures. 
Why not use proper locks instead? A semaphore or countdown-latch would be the perfect primitive to use here. (@see AsyncJobFileStagingHandlerTest)

EgorZhuk:
@tmorgner please take look

test-src/org/pentaho/reporting/platform/plugin/cache/FileSystemCacheBackendTest.java
```diff
@@ -135,7 +148,7 @@ public String getValue() {
 
     private void writeObject( final ObjectOutputStream oos ) throws IOException {
       try {
-        Thread.sleep( 3000L );
+        Thread.sleep( 1000L );
```
tmorgner: There is no need to do a thread-sleep. If you want to enforce a certain order of operations, do so via latches. This way you wont slow down the unit tests.

EgorZhuk: @tmorgner it's not enforcing order, it's mocking of writing/reading real objects that usually take some time, otherwise file locks will be released immediately

tmorgner: That is what locks are for. Use two locks/latches for enforcing the right order. 

For instance for "testWriteWhileReading" you can enforce the order by setting and resetting latches while you are in readObject(..). These latches must be static, or you could not get a reference to them, but otherwise it is just standard code:

```Java

  @Test
  public void testWriteWhileReading() throws Exception {
    final String val = "ShinySecretValue";
    final String val2 = "ShinySecretValue2";
    final ExecutorService executor = Executors.newFixedThreadPool( 2 );
    readStartedSignal = null;
    writeStartedSignal = null;
    // setup writes without any signals ...
    fileSystemCacheBackend.write( Arrays.asList( directoryKey, key ), new LongRead( val ) , new HashMap<String, Serializable>());
    // now we can setup the signals ..
    readStartedSignal = new CountDownLatch(1);
    writeStartedSignal = new CountDownLatch(1);

    final Future<LongRead> read = executor.submit( new Callable<LongRead>() {
      @Override public LongRead call() throws Exception {
        return (LongRead) fileSystemCacheBackend.read( Arrays.asList( directoryKey, key ) );
      }
    } );
    final Future<Boolean> write = executor.submit( new Callable<Boolean>() {
      @Override public Boolean call() throws Exception {
        readStartedSignal.await();
        // unblock the reader ...
        writeStartedSignal.countDown();
        return fileSystemCacheBackend.write( Arrays.asList( directoryKey, key ), new LongRead( val2 ), new HashMap<String, Serializable>() );
      }
    } );
    assertTrue( write.get() );
    assertEquals( read.get().getValue(), val );
  }

  @After
  public void clearLatchesAfterEachTest() {
    readStartedSignal = null;
    writeStartedSignal = null;
  }

  private static CountDownLatch readStartedSignal;
  private static CountDownLatch writeStartedSignal;


  private class LongRead implements Serializable {

    private String value;

    public LongRead( final String value ) {
      this.value = value;
    }

    public String getValue() {
      return value;
    }

    private void writeObject( final ObjectOutputStream oos ) throws IOException {
      oos.writeObject( value );
    }

    private void readObject( final ObjectInputStream ois ) throws ClassNotFoundException, IOException {
      try {
        if ( readStartedSignal != null ) {
          // unlock the writer ...
          readStartedSignal.countDown();
        }
        if ( writeStartedSignal != null ) {
          // wait for the writer to start working ..
          writeStartedSignal.await();
        }
        // now that small time frame should be enough, as we guarantee that the writer is
        // ready to enter the file-system write method as soon as it sees the latch go.
        Thread.sleep(100);
        final Object o = ois.readObject();
        this.value = (String) o;
      } catch ( final InterruptedException e ) {
        e.printStackTrace();
      }

    }
  }
```

EgorZhuk:
@tmorgner Thank you, now I see. Applied the same to ReadWhileWriting

PR #316, opened 2016-04-13
[BACKLOG-7605] Glass pane needed when loading first page

Aliaksandr-Kastenka:
@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -787,6 +787,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         var isFirstContAvStatus = true;
         var isIframeContentSet = false;
         var isFinished = false;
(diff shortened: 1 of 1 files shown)


package-res/reportviewer/reportviewer.js
```diff
@@ -1098,6 +1099,15 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         });
       },
 
+      _hideDialogAndPane: function() {
+        var dlg = registry.byId('feedbackScreen');
+        dlg.hide();
+        if(this._updateReportTimeout >= 0) {
+          this._updateReportTimeout = -1;
```
tmorgner: If you remove the handle you got from the "setTimeOut" function call, does that mean you want to cancel that timeout? In that case, wouldn't you want to call "clearTimeOut(..)" as well?

Aliaksandr-Kastenka:
@tmorgner 
Pull-request updated.

PR #318, opened 2016-04-18
[BACKLOG-7598] - Fix chrome pdf viewver plugin bogus behaviour

prakapenka:
This brings next fixes for us:
1) We need a special endpoint for pdf viewer plugin now able to send get
request to whatever it want it to send. Since we can't fix chrome, we have
to fix JobManager fot that.

2) We need to keep same content between requests.
So since we expected to keep content we don't clean-up future after
content requested. Instead - we keep a collection of futures. We expect
clean-up when session destroyed. Also content is kept in staging folder.

3) Since we keep content - we introduce new methods for
IFixedSizeStreamingContent.cleanContent() to explicitly clean up content.

And yes - finally we came to a one-more cahce layer - all content is kept in
statging folder during server works.

Pay attention - content for all currently running sessions.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/JobManager.java
+++ b/src/org/pentaho/reporting/platform/plugin/JobManager.java
@@ -80,6 +80,13 @@ public class JobManager {
     return null;
   }
 
+  @GET @Path( "{job_id}/content" )
+  public Response getPDFContent( @PathParam( "job_id" ) String job_id ) throws IOException {
+    logger.debug( "Chrome pdf viewer workaround. See BACKLOG-7598 fir details" );
+
+    return this.getContent( job_id );
+  }
+
   @POST @Path( "{job_id}/content" ) public Response getContent( @PathParam( "job_id" ) String job_id )
     throws IOException {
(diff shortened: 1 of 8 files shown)


prakapenka:
@tmorgner please review

src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -80,6 +80,13 @@ public JobManager( final boolean isSupportAsync, final long pollingIntervalMilli
     return null;
   }
 
+  @GET @Path( "{job_id}/content" )
```
tmorgner: Given that the method is the same (except for the logging) as the original getContent method, simply add the GET annotation to the other method. A REST endpoint can have multiple HTTP-method annotations.

This then eliminates this method and we dont have to write unit tests for another method.

prakapenka:
"multiple HTTP-method annotations" - will compile green but then fail on run-time

tmorgner:
OK, if the REST API wants to be picky, then so be it. Let it be merged.

PR #321, opened 2016-04-25
[BACKLOG-6745] - enchase perfomance logging with additional info

prakapenka:
1) Introduce PluginPerfomanceMonitorContext that will provide session info
for reporting perfomance logging.
We will track perfomance metrics with logging table records with
instanceId from tomcat controller and 'message' field in audit log.

2) Re-check audit logging, remove static hard-coded invocation to be able
to mock in junits.

See original jira case for details.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/AuditWrapper.java
+++ b/src/org/pentaho/reporting/platform/plugin/AuditWrapper.java
@@ -1,0 +1,48 @@
+/*
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License, version 2 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/gpl-2.0.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ *
+ * Copyright 2006 - 2016 Pentaho Corporation.  All rights reserved.
+ */
+
+package org.pentaho.reporting.platform.plugin;
+
+import org.pentaho.platform.api.engine.ILogger;
+import org.pentaho.platform.engine.core.audit.AuditHelper;
+
+/**
+ * Wrapper to a static AuditHelper call. Simplify mocking and get rid of static dependencies in main code.
+ */
+public class AuditWrapper {
+
+  public static final AuditWrapper NULL = new NullAudtitWrapper();
+
+  public void audit( String instanceId, final String userId, String actionName, final String objectType,
+                     String processId, final String messageType, final String message, final String value, final float duration,
+                     final ILogger logger ) {
+    // register execution attempt
+    AuditHelper.audit( instanceId, userId, actionName, objectType, processId,
+      messageType, message, value, duration, logger );
+  }
+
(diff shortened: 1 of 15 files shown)


prakapenka:
@tmorgner please review

src/org/pentaho/reporting/platform/plugin/AuditWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.pentaho.platform.api.engine.ILogger;
+import org.pentaho.platform.engine.core.audit.AuditHelper;
+
+/**
+ * Wrapper to a static AuditHelper call. Simplify mocking and get rid of static dependencies in main code.
+ */
+public class AuditWrapper {
+
+  public static final AuditWrapper NULL = new NulAudtitWrapper();
```
tmorgner: Null has two Ls

src/org/pentaho/reporting/platform/plugin/PluginPerfomanceMonitorContext.java
```diff
@@ -0,0 +1,72 @@
...
+public class PluginPerfomanceMonitorContext implements PerformanceMonitorContext {
+
+  public static final String FORMAT = " sessionid[%1s] userid[%2s] requestid[%3s]";
+
+  @Override public PerformanceLoggingStopWatch createStopWatch( String tag ) {
+    return new LoggingStopWatch( tag, this.getMessage() );
+  }
+
+  @Override public PerformanceLoggingStopWatch createStopWatch( String tag, Object message ) {
+    return new LoggingStopWatch( tag, message + this.getMessage() );
```
tmorgner: Instead of using string + string, pass the "message" object into the "getMessage" method and format everything in one operation. 

Your format string would then be "%s sessionid[%s] userid[%s] requestid[%s]"

You probably dont intend to use a one-space padding, two-space padding and 3 space padding on the format string either.

test-src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncExecutionAuditTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    PentahoAsyncReportExecution execution =
+      new PentahoAsyncReportExecution( url, component, handler, session, expected, wrapper );
+    //execution.notifyTaskQueued( uuid );
+
+    PentahoAsyncExecutor executor = new PentahoAsyncExecutor( 2 );
+    executor.addTask( execution, session );
+
+    // hope 10 sec is enough
+    latch.await( 10, TimeUnit.DAYS );
```
tmorgner: Do you really want to wait 10 days?

src/org/pentaho/reporting/platform/plugin/ExecuteReportContentHandler.java
```diff
@@ -60,9 +61,12 @@ public ExecuteReportContentHandler( final ReportContentGenerator contentGenerato
 
   public void createReportContent( final OutputStream outputStream, final Serializable fileId, final String path,
       final boolean forceDefaultOutputTarget ) throws Exception {
+    ReportListenerThreadHolder.setRequestId( contentGenerator.getInstanceId() );
+
     final long start = System.currentTimeMillis();
     final Map<String, Object> inputs = contentGenerator.createInputs();
 
+    // Deprecated call. See AuditWrapper.
```
tmorgner: Can we use the new AuditWrapper here as well? I cannot see a good reason why we can't use the cleaner method everywhere. It does the same thing as before, so the behaviour is not going to change, but at least we dont end up with a mix of old and new code.

tmorgner:
Apart from the one comment, it looks sane and mergable.

prakapenka:
refactor ExecuteReportContentHandler, remove static methods dependencies, fix audit records. If report failed to validate - we treat this as FAIL in audit records (previously this was tracked as success).

Added junit for createReportContent method and for success/failed responses.

@tmorgner please review.

tmorgner:
Nice catch with the wrong audit-status. Merged.

PR #322, opened 2016-04-27
Backlog 7929

EgorZhuk:
Hello @tmorgner, I have a couple of thingth for review:

1. Clean futures and tasks in a proper way
2. Move auto-scheduling to server side
3. Implement simple UI to view large paginated reports without hanging a browser (see video)

https://drive.google.com/file/d/0BzVckAbDytTXNU5FOXNqODJFVWM/view

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
+++ b/src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
@@ -18,10 +18,15 @@
 
 package org.pentaho.reporting.platform.plugin.async;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.pentaho.reporting.engine.classic.core.ClassicEngineBoot;
 import org.pentaho.reporting.engine.classic.core.event.ReportProgressEvent;
+import org.pentaho.reporting.engine.classic.core.event.ReportProgressListener;
 import org.pentaho.reporting.libraries.base.config.ExtendedConfiguration;
+import org.pentaho.reporting.libraries.base.util.ArgumentNullException;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.UUID;
 
 /**
@@ -38,6 +43,7 @@ public class AsyncReportStatusListener implements IAsyncReportListener {
   private final String path;
   private final UUID uuid;
   private final String mimeType;
+  private final List<? extends ReportProgressListener> callbackListeners;
   private String errorMessage = null;
   private AsyncExecutionStatus status = AsyncExecutionStatus.QUEUED;
   private int progress = 0;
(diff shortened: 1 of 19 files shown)


test-src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncExecutionTest.java
```diff
@@ -117,7 +120,7 @@ public void testListenerFailExecution() throws Exception {
 
     PentahoAsyncReportExecution exec = createMockCallable();
     // simulates queuing
-    exec.notifyTaskQueued( UUID.randomUUID() );
+    exec.notifyTaskQueued( UUID.randomUUID(), null );
```
tmorgner: This call should actually error out. If you have a list parameter, never pass in Null, demand that your API user passes in Collections.emptyList() instead.

src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListener.java
```diff
@@ -52,11 +56,13 @@
...
   public AsyncReportStatusListener( final String path,
-                             final UUID uuid,
-                             final String mimeType ) {
+                                    final UUID uuid,
+                                    final String mimeType,
+                                    final List<? extends ReportProgressListener> callbackListeners ) {
     this.path = path;
     this.uuid = uuid;
     this.mimeType = mimeType;
+    this.callbackListeners = callbackListeners;
```
tmorgner: Always check your list parameters for null. Dont allow it, demand that they pass in an empty list instead. Then you dont have to check for null all the time.

Do you expect that users are allowed to modify that list after it has been set? If not, make a copy for your private use to isolate yourself from evil user actions.

EgorZhuk:
Another thing we need here - replace CSS links with content to make it work when user downloads the file.

tmorgner:
There are conflicts. Can you rebase this agains the latest EWOK branch?

EgorZhuk:
@tmorgner done. Also bukk check added. I'm working on embedded css and images now.

test-src/org/pentaho/reporting/platform/plugin/async/AsyncReportStatusListenerTest.java
```diff
@@ -1,7 +1,9 @@
 package org.pentaho.reporting.platform.plugin.async;
```
tmorgner: No license header.

src/org/pentaho/reporting/platform/plugin/output/PaginationControlWrapper.java
```diff
@@ -0,0 +1,77 @@
+package org.pentaho.reporting.platform.plugin.output;
```
tmorgner: No license header.

tmorgner:
Only some license headers missing, then I can merge.

EgorZhuk:
@tmorgner added licenses, also added code to embed css. It works fine with styles but the same trick with images it too slow :(

EgorZhuk:
@tmorgner I just checked how platform scheduler works - it places css and image in export folder and replaces links wih "api/repo/foo". I believe it's not what we want because we will have css for every page and numerous images.

EgorZhuk:
@tmorgner

EgorZhuk:
@tmorgner I see two ways of making images work:

1) Do it like in a platform scheduler - it will require basic auth if user wants to see pictures
2) Do it well - add images to the archive that user downloads to his machine

PR #328, opened 2016-04-29
[BACKLOG-7916] Error message 'Request failed' appears when opening re…

Aliaksandr-Kastenka:
…ports with mandatory prompts

@tmorgner 
Please review.


EgorZhuk:
@Aliaksandr-Kastenka please rebase your commits with latest EWOK

tmorgner:
This is only half of the story. If the prompts are not valid, you should not even get to this point. The report-viewer should check the status of the prompts to see whether there are any errors, and should only send the request to the server when everything is fine. 

You can get that state via the prompt-api via 

        var state = prompt.api.operation.state();
        var parametersValidFlag = state.promptNeeded;

Aliaksandr-Kastenka:
https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/329

PR #333, opened 2016-05-03
[BACKLOG-7993] - Don't close stream you didn't open

EgorZhuk:
"Error occurs on attempt to open or create a PIR report if CachePageableHtmlContent=false"
Something is closing the stream 0__0
@tmorgner please take a look

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/output/PageableHTMLOutput.java
+++ b/src/org/pentaho/reporting/platform/plugin/output/PageableHTMLOutput.java
@@ -20,6 +20,7 @@ package org.pentaho.reporting.platform.plugin.output;
 import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.ByteArrayOutputStream;
 
 import org.pentaho.platform.api.engine.IApplicationContext;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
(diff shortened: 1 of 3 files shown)


tmorgner:
Can you find out why the heck that stream is closed? A subclassing of your byte-array stream and the logging of an exception on close should do the trick. I'd rather fix the cause of the problem than the symptom. If it turns out that the stream is closed by code we dont control (unlikely) we can use this work-around. But otherwise I'd rather fix the real beast.

EgorZhuk:
@tmorgner it's in engine [line](https://github.com/pentaho/pentaho-reporting/blob/master/engine/core/source/org/pentaho/reporting/engine/classic/core/modules/output/table/html/helper/AbstractHtmlPrinter.java#L373) we are closing the writer and it closes the stream we passed in ` this.writer = createWriterService( documentContentItem.getOutputStream() );`
       ` xmlWriter = writer.getXmlWriter();`

tmorgner:
Yes. And tell the author of that piece of code that closing streams randomly is naughty. Seriously, that guy needs to learn how to program safely! ;)

Use a "writer.flush()" instead just to make sure that nothing is left in some random buffer.

tmorgner:
Ahh, just looking at the code. The close exists because we may acquire a new stream from the "documentContentItem" (see HtmlPrinter line 248). 

So the proper fix would be one layer deeper: In StreamRepository (in module librepository), we have a WrappedInput- and -OutputStream that should not call close on the parent. 

This way, if you are working against files, you'll close them correctly (uses "FileRepository" or "ZipRepository"), but if you work against streams, you keep the original streams open.

PR #336, opened 2016-05-05
[BACKLOG-7898] - log cancelled status for audit table

prakapenka:
- Add cancelled status for cancelled executions
- Fix excessive handler 'handler.getStagingOutputStream()' : we don't need to close it in catch block since we do close it in finally block
- Added tests for execution cancellations in thread pool as detailed as possible.


prakapenka:
@tmorgner please review

PR #341, opened 2016-05-06
[BACKLOG-8074] - Cancel old job if output type changed

EgorZhuk:
@tmorgner please review


EgorZhuk:
also fixes BACKLOG-8101

PR #346, opened 2016-05-13
[CLEANUP] - Make async logic in JS more documented and clean

EgorZhuk:
Hello Thomas,

We faced common JS issue when you fix something and another part breaks ( may be we need to write JS tests too?).
I refactored async logic in js to make it more maintainable, also added comments to make it more clear what is happening.

While doing it I also fixed a couple of bugs :)

[BACKLOG-8100] - Page loading notification is displayed after cancellation
of report
[BACKLOG-8072] - Page caching counter displays above glass pane
if requested not-cached page
[BACKLOG-8097] - Report is downloaded twice
when changing parameters of report
[BACKLOG-8098] - Report is downloaded
when changing output type though Auto-submit is unchecked

I tested it manually and we will have a couple of days before merge to master for QA team to test it too.

Please take a look.

@tmorgner

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -737,10 +737,10 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
               this.reportPrompt.mode = 'MANUAL';
               this.reportPrompt.api.operation.refreshPrompt();
             }else{
-              this._updateReportContentCore();
+              this._updateReportContent();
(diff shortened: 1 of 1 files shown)


tmorgner:
Looks good to me. There are some JavaScript tests in the project already, so you should be able to run "ant test-js" to invoke the Karma test runner. However, those tests are not run by either Wingman or the CI server, I believe. But they should make the JS-test procedure a bit easier and/or repeatable.

PR #347, opened 2016-05-13
[BACKLOG-8070] Report page and page caching counter disappear when as…

Aliaksandr-Kastenka:
…k not-cached page

@EgorZhuk @tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -1119,7 +1119,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
           clearTimeout(this._updateReportTimeout);
           this._updateReportTimeout = -1;
 
-          if(isTimeout) {
+          if(isTimeout && !this.reportPrompt._isReportHtmlPagebleOutputFormat) {
             // This happens, specifically, when the user selects a downloadable output format.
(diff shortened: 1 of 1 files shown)


tmorgner:
I am not quite sure about this fix here. The "_submitReportEnded" function is called from many places to clean up once we have content that should be displayed or we no longer wait for status updates (ie status is cancelled, failed or scheduled).

However, it is also called from the very start of the _updateReportContentCore function in what seems to be a workaround or fix for case PRD-3962. I am pretty sure that your bug is caused by this particular invokation (Line 838). 

I think in the context of our async mode handling, this call is actually harmful and should be disabled whenever async-mode is enabled. Otherwise it will kill the glasspane as soon as 5 seconds have passed, regardless of whether there is content available or not.

Our async-callback actually handles the glasspane and calls "_submitReportEnded" whenever a final state has been reached. This handling is independent of the iframe as we monitor the job-state directly. 

@EgorZhuk you have been doing a great job at cleaning up the logic there and your brain is probably filled with all the knowledge about it right now - what do you think of that callback and the fix?

EgorZhuk:
@tmorgner I have a simple answer - it's not reproducible after the refactoring :)
@Aliaksandr-Kastenka thank you for the fix, you can close the PR

PR #351, opened 2016-05-19
[BACKLOG-8361] - Unable to navigate to the last page of saved in HTML report

EgorZhuk:
@tmorgner please take a look

### DIFF
--- a/package-res/reportviewer/paginationTemplate.html
+++ b/package-res/reportviewer/paginationTemplate.html
@@ -34,8 +34,8 @@
 <div>
   <div class="page-control">
     <div class="page-control-nav">
-      <span id="left" onclick="left()"><&nbsp;</span>
-      <span id="right" onclick="right()">>&nbsp;</span>
+      <span id="left">&lt;&nbsp;</span>
+      <span id="right">&gt;&nbsp;</span>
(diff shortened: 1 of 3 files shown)


EgorZhuk:
Don't lie Wingman :)

EgorZhuk:
lol :+1:

package-res/reportviewer/paginationTemplate.html
```diff
@@ -34,8 +34,8 @@
 <div>
   <div class="page-control">
     <div class="page-control-nav">
-      <span id="left" onclick="left()"><&nbsp;</span>
-      <span id="right" onclick="right()">>&nbsp;</span>
+      <span id="left"><&nbsp;</span>
+      <span id="right">>&nbsp;</span>
```
tmorgner: Nope, that line does not look correct. There are two >> characters on the right-span.

tmorgner: Oh, I see what is going on. Please encode the > and < characters, so that HTML parsers are happy. And literal > or < should be encoded as &lt; and &gt;

EgorZhuk:
@tmorgner github for sure knows how to encode it :smile:

PR #353, opened 2016-05-19
[BACKLOG-8395] - Reports saved in HTML (paginated) are broken if mult…

EgorZhuk:
…iple reports run at the same time

Stupid me ;)

@tmorgner please review


EgorZhuk:
@tmorgner I belive somebody have broken wingman, it is extra-wierd today :(

PR #356, opened 2016-05-25
[CLEANUP] - Cleanup cache files when platform stops

EgorZhuk:
We have several cases when cache files may not be cleaned up properly:

- cache for scheduled reports is not purged because we are already out of session scope
- errors may occur when trying to purge cache on logout

Let's add extra place to collect garbage - on platform stop.
Session cache is purget totally, time-based cache cleans only old files.

@tmorgner please take a look

### DIFF
--- a/package-res/reportviewer/dojo/pentaho/reportviewer/FeedbackScreen.html
+++ b/package-res/reportviewer/dojo/pentaho/reportviewer/FeedbackScreen.html
@@ -1,9 +1,9 @@
 <table class="feedbackscreen-content" onkeyup="">
     <tr>
-        <td>
+        <td class="feedbackscreencontentbody">
             <table class="upper-table">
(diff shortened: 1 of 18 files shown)


EgorZhuk:
@tmorgner it seems that we have floating coverage from one test run to another - look, coverage increased for all changed files but suddenly decreased for untouched ones. Should I investigate it?

EgorZhuk:
@tmorgner also it may be a good place too kill css/images too

tmorgner:
Floating coverage normally means that such code is only indirectly tested. It is simply ran through while other tests were running, and thus it counts as "covered" even when no assertions were made.

It is probably a very good idea to explicitly cover the behaviour of these "floaters" so that changes in unrelated tests do not affect them anymore.

PR #357, opened 2016-05-25
[SP-2684][BISERVER-13171] OK button is disabled even if all mandatory…

mchen-len-son:
… settings are specified when scheduling PIR

http://jira.pentaho.com/browse/BISERVER-13171
added afterRender logic


mchen-len-son:
@tmorgner @pamval 

please review. 
I don't have access to wingman, and canot therefore see what happened there.
Probably it's not configured yet for this project, but if it's my fault, please let me know.

This is a backport of #348.

PR #358, opened 2016-05-26
[CLEANUP] - Cleanup cache files when platform stops

EgorZhuk:
Plus tests


EgorZhuk:
@tmorgner We are ready to go, I've just extracted PentahoAsyncExecutor.AutoScheduleListener to  AutoScheduleListener

tmorgner:
Can you give the auto-schedule listener a little bit more love? It is down by 60% on coverage and probably was the culprit for the changing numbers all along. At least the refactored class is easy enough to allow tests with mocks now.

EgorZhuk:
@tmorgner it is 100% covered by unit tests, I will add some integration tests

PR #359, opened 2016-05-27
[BACKLOG-8417] Fix zero length report when > 1 logout event

EgorZhuk:
Please don't merge.


EgorZhuk:
Optimized test memory usage

PR #360, opened 2016-05-27
WINGMAN TEST

EgorZhuk:
Please don't merge.


EgorZhuk:
@tmorgner I totally like this metal guy

tmorgner:
@wingman-pentaho is a miserable character and misery needs company. 

And after all, you did not unit-test your added comment!!!! How can you prove that you commented in the correct spot without proper tests.

PR #361, opened 2016-05-27
[BACKLOG-8579] - Fix scheduling for partially saved reports

EgorZhuk:
Fix scheduling for partially saved reports


EgorZhuk:
@wingman-pentaho I summon you!

PR #362, opened 2016-05-27
Bugfixes, refactoring of inner classes, code coverage

EgorZhuk:
Let's see Wingman results.


tmorgner:
File /home/buildguy/project/test-src/org/pentaho/reporting/platform/plugin/ReportingActionIT.java

    Message: Using a static member import should be avoided - org.pentaho.di.core.util.Assert.assertTrue. at Line: 37


I think the import for that is wrong. There is a perfectly usable JUnit assert somewhere on the classpath.

EgorZhuk:
@tmorgner Could you please share wingman's log, I don't understand why some tests are failing. I tried to comment out all new tests and nothing changed. Also I tried both fork and non-fork mode on local machine and all tests passed.

tmorgner:
Seems like Jacoco messes things up here. Nothing we can do, ping the build team to let them look into it. 

> Mon May 30 02:31:06 EDT 2016 - stdout - head:     [junit] Error opening zip file or JAR manifest missing : /tmp/jacocoagent8497557208141832351.jar
> Mon May 30 02:31:07 EDT 2016 - stdout - head:     [junit] Running org.pentaho.reporting.platform.plugin.async.AsyncSystemListenerIT
> Mon May 30 02:31:07 EDT 2016 - stdout - head:     [junit] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0 sec
> Mon May 30 02:31:07 EDT 2016 - stdout - head:     [junit] Test org.pentaho.reporting.platform.plugin.async.AsyncSystemListenerIT FAILED (crashed)
> Mon May 30 02:31:07 EDT 2016 - stdout - head:     [junit] Error opening zip file or JAR manifest missing : /tmp/jacocoagent8497557208141832351.jar
> Mon May 30 02:31:07 EDT 2016 - stdout - head:     [junit] Error occurred during initialization of VM
Mon May 30 02:31:07 EDT 2016 - stdout - head:     [junit] agent library failed to init: instrument

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -202,10 +202,18 @@ public synchronized int generate( final MasterReport report, final int acceptedP
       }
 
 
-      final IReportContent data = regenerateCache( report, yieldRate, key, acceptedPage );
-      outputStream.write( data.getPageData( acceptedPage ) );
+      final IReportContent fullReportCache = regenerateCache( report, yieldRate, key, acceptedPage );
```
EgorZhuk: The fix is here.

src/org/pentaho/reporting/platform/plugin/staging/AsyncJobFileStagingHandler.java
```diff
@@ -119,9 +124,12 @@ public static void cleanSession( final IPentahoSession iPentahoSession ) {
 
     // some lib can do it for me?
     try {
-      FileUtils.deleteDirectory( sessionStagingContent );
+      if ( sessionStagingContent != null && sessionStagingContent.isDirectory()
```
EgorZhuk: The fix is here.

src/org/pentaho/reporting/platform/plugin/async/PersistReportToJcrTask.java
```diff
@@ -89,14 +91,25 @@
 
       int copy = 1;
 
-      while ( dataLocation.exists( targetName ) ) {
-        targetName = String.format( FORMAT, cleanFileName, copy, targetExt );
-        copy++;
+      final OutputStream outputStream;
```
EgorZhuk: The fix is here.

src/org/pentaho/reporting/platform/plugin/async/PersistReportToJcrTask.java
```diff
@@ -89,14 +91,25 @@
...
+
+      synchronized ( FORMAT ) {
```
tmorgner: Instead of using a static lock (JVM wide lock-downs are sooo 20th century!), can you lock on the Repository-File instead?

You can use org.pentaho.platform.repository2.unified.jcr.ILockHelper's lockFile and unlockFile method for that. You get an instance of that class via the PentahoSystem#get method, as usual. Use "outputFolder.getId()" for as argument for the locking.

According to the documentation on DefaultLockhelper, this will create a deep lock (covers the node and all descendent nodes). 

Be aware that the lock will survive the session, so a try-finally is mandatory.

EgorZhuk:
@tmorgner for ILockHelper we need jcr.Session. It is stored internally in PentahoJcrTemplate. We can try to use template but we will need to add dependency for org.springframework.extensions.jcr. to make it work. The only place where this helper is used is JcrTempalte itself, so I'm not sure if it is suitable for us.

tmorgner:
Ha, we may produce sometimes creepy but useful internal APIs that need bad hacks to be accessible, but luckily locking is not one. I just had to look a bit further up the food chain for it.

Use org.pentaho.platform.api.repository2.unified.IUnifiedRepository#lockFile instead. You can get the IUnifiedRepository from PentahoSystem as well, and now all you need is the file-Id and a text message telling the system why you want to lock it (so that any blame for deadlocks can blamed on you - isn't the API nice?).

EgorZhuk:
@tmorgner unfortunately it's not that easy :) You can't lock on directory because it's not a lockable mixin type. We need to lock on file, and the file can be not created yet ( actually it's desired condition ), so we need to create new one and deal with java synchronization again. If we go to dataLocation.exists and createItem we will see that it does exactly the same. 

If we want more granular lock we can synchronize on session object but we can have issues with some  scheduling directory strategies which share directory between users.

EgorZhuk:
@tmorgner FYI: I've checked synchronization timing - it's 300-500 ms on 9 reports started at he same time ( scheduled all on largest dashboard)

tmorgner:
Yeah, if the locking code in the platform does not want us to lock on directories, then screw it. Global lock it is for now. This may backfire in a clustered setting, but I feel we run out of options here. 

And I see that Jacoco is back to haunt us. Nice.

tmorgner:
I feel conflicted now, after merging the other pull request #366 . Can this be closed? And if not, can we build a Jacoco-agent-jar voodoo doll and stick needles into it until the tests pass?

EgorZhuk:
Separate pull requests will be opened for each commit.

PR #363, opened 2016-05-30
[BACKLOG-8423] Cancel request is sent only once if dashboard mode is …

Aliaksandr-Kastenka:
…changed

Instead of window.IsSwitchingToEditMode, I am checking pho.dashboards.editMode, similar to how it was done in dashboard plugin:
https://github.com/pentaho/pentaho-platform-plugin-dashboards/blob/master/resource/script/pentaho-dashboard-controller.js#L1810-L1813

@tmorgner 
Please review.


tmorgner:
Sweet. Hooking on to the dashboard mode and the flag to produce the same result without having to change the dashboard-core - I love it!

PR #366, opened 2016-06-01
[BACKLOG-8579] - Fix scheduling for partially saved reports

EgorZhuk:
Separate PR for 8579 with all tests.


EgorZhuk:
@tmorgner it's the 1st commit from #362 and jacoco doesnt' fail here - we can try to merge them one by one and find the issue. Coverage in InternalOutputHandlerSelector compalins about getter in private inner class that returns null :+1: https://yadi.sk/i/_Yxz_0WSsB5jg

tmorgner:
Lets try that.

PR #367, opened 2016-06-02
[BACKLOG-8417] - Fix zero length report when > 1 logout event

EgorZhuk:
The second one from #362 
@tmorgner


EgorZhuk:
Yeah, so this is the one that brakes Jacoco. I will play with it.

EgorZhuk:
@tmorgner so the deal is not in tests, i've commented them, it's something else

PR #368, opened 2016-06-02
[CLEANUP] - Update IDEA project JDK to 1.8

EgorZhuk:
Just in order not to switch it each time as far as sources are already 1.8

@tmorgner

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -3,8 +3,7 @@
   <component name="EntryPointsManager">
     <entry_points version="2.0" />
   </component>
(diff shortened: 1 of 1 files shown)


tmorgner:
Seriously. Again some lower test coverage. @wingman-pentaho does not like you!

Merged anyway, as Wingman is a bit of a .. .. ... friendly helper.

lucboudreau:
Coverage moved from unit to integration, that's what your friendly Wingman buddy is telling you. :)

PR #369, opened 2016-06-02
[BACKLOG-8417] - Fix zero length report when > 1 logout event

EgorZhuk:


tmorgner:
Hey look, org.pentaho.reporting.platform.plugin.async.PentahoAsyncExecutor.WriteToJcrTask no longer exists, but has only lost 33% of its method coverage. Which means that this code is now officially a ghost in the shell.

EgorZhuk:
@tmorgner magic  same changes work now :+1:  I will add some more junits.

PR #372, opened 2016-06-06
[CLEANUP] - Fix naughty test

EgorZhuk:
@tmorgner please take a look. Just replaced NullInputStream and increased concurrency level.


tmorgner:
Is there any reason why 100 iterations are more validating than 10 iterations?

EgorZhuk:
@tmorgner I will rewrite it to 100% meet race condition

PR #374, opened 2016-06-08
[BACKLOG-8752] - Enhance persistance logic to deal with user-selected location

EgorZhuk:
The idea of this change is simle - provide an API endpoint to change target location of the scheduled report task.

- prompting is configurable in spring
- scheduling functionality is mostly unchanged
- we move file from default scheduling directory to provided target
- the case when report is already written and the we whant to cahnge the location is handled by the second listener which simpy memorize all scheduling locations
- all listeners but memorize one are removed after they were notified
- api deals with file ids but could be easily changed to paths

Further enchancements:
- provide scheduling strategy that deals with tmp dir
- provide mechanism to remove file from repo if location prompt was closed

@tmorgner Please take a look


EgorZhuk:
Ah, and the test with jcr race condition is rewritten.

src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -64,11 +64,21 @@ public JobManager() {
...
   public JobManager( final boolean isSupportAsync, final long pollingIntervalMilliseconds,
+                     final long dialogThresholdMillisecond, final boolean promptForLocation ) {
+    if ( !isSupportAsync ) {
+      logger.info( ASYNC_DISABLED );
+    }
+    this.config = new Config( isSupportAsync, pollingIntervalMilliseconds, dialogThresholdMillisecond,
+      promptForLocation );
+  }
+
+  public JobManager( final boolean isSupportAsync, final long pollingIntervalMilliseconds,
```
tmorgner: Avoid duplicate code (and magically cover more lines in tests) by re-routing this constructor to the one with 4 parameters.

src/org/pentaho/reporting/platform/plugin/JobManager.java
```diff
@@ -237,6 +247,24 @@ public Response schedule( @PathParam( "job_id" ) final String jobId ) {
...
+    final IPentahoSession session = PentahoSessionHolder.getSession();
+    final UUID uuid;
+    try {
+      uuid = UUID.fromString( jobId );
+    } catch ( final Exception e ) {
+      logger.error( INVALID_UUID + jobId );
+      return get404();
+    }
+
+    executor.updateSchedulingLocation( uuid, session, folderId );
```
tmorgner: You never check whether the jobId is actually valid. I can spam you with infinite syntactically correct UUIDs and you will happily fill up the location map. (The executor implementation does not check the UUID either. 

Second: REST principles state that if you want to modify state on the server, you never ever use a GET method for that. GETs may be cached by a proxy, and are generally assumed to be read only operations. 

I would also say that the folder ID should probably not be a path parameter. In the context of this operation is does not make much sense to present the folders as tree/path structure. Use a normal form parameter instead.

/{job_id}/updateLocation?folderId=/my/path/here.

If you really want to use a path parameter, I would swap folderId and the "updateLocation" key, as "updateLocation" is an operation on the job, not on the path.

EgorZhuk:
@tmorgner fixed code review issues

EgorZhuk:
Fixed checkstyle

EgorZhuk:
@tmorgner could you please take a look

tmorgner:
Just a few more targetted tests to bring up the code coverage and it can be happily merged.

EgorZhuk:
More tests for the lord of tests.

lucboudreau:
Please make sure to squash these commits before merging. Thanks!

EgorZhuk:
Ok, let's do it again.

EgorZhuk:
@tmorgner Added newName query param as far as user is able to change the name from the location prompt. Also added file name calculation code ( to avoid errors for same file names ).

EgorZhuk:
@tmorgner I looked at this 1% in uit test for executor and found out that the cleanup code is never called. As far as we cancelling task onSuccess is not reached. In case of scheduling files are not removed too. So I've added cleanup in proper places (on failure, on cancel, after scheduling callback) and wrote tests for each one. 

About integration tests coverage should I work on it? 

- CompositeKey is fully covered besides equals, I believe it's not a case for integration tests?
- We didn't touch ReportContentLocation, so it's just because of using mocks.
- May be some work  could be done for executor and job manager.

tmorgner:
The output is a bit confusing, I agree. I care most about the unit test coverage. There are no sensible integration tests we can do (at our module level) with JAX-RS classes like the JobManager (or deep framework classes like ReportContentLocation), so it is OK to not have integration-test coverage for it for as long we have unit-test coverage.

PR #377, opened 2016-06-16
BACKLOG-8976 - Async jobs fail with 404 in PIR

EgorZhuk:
- use polymorfic getExecutor method to get executor from the desired context

@tmorgner this kind of bugs could only be cacthed before manual testing if we have integration tests for the whole platform :(


EgorZhuk:
It seems that we have autotests for such case but nobody cares to report bugs if they fails.

PR #378, opened 2016-06-21
[BACKLOG-8501] - PIR: Sometimes message "The row limit has been reach…

dzmitry-hryhoryeu:
[BACKLOG-8501] - PIR: Sometimes message "The row limit has been reached" disappears when row limit is applied	

@tmorgner 
Please review.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/cache/AbstractReportContentCache.java
+++ b/src/org/pentaho/reporting/platform/plugin/cache/AbstractReportContentCache.java
@@ -20,6 +20,7 @@ package org.pentaho.reporting.platform.plugin.cache;
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Abstract implementation of caching policy
@@ -37,12 +38,21 @@ public abstract class AbstractReportContentCache implements IReportContentCache 
 
   @Override
   public boolean put( final String key, final IReportContent value ) {
-    return getBackend().write( computeKey( key ), value, new HashMap<String, Serializable>() );
+    return put( key, value, new HashMap<String, Serializable>() );
+  }
+
+  @Override
+  public boolean put( final String key, final IReportContent value, Map<String, Serializable> metaData ) {
+    return getBackend().write( computeKey( key ), value, metaData );
   }
 
   @Override
   public IReportContent get( final String key ) {
     return (IReportContent) getBackend().read( computeKey( key ) );
+  }
+
+  @Override public Map<String, Serializable> getMetaData( String key ) {
+    return getBackend().readMetaData( computeKey( key ) );
   }
 
   public ICacheBackend getBackend() {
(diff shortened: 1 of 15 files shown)


src/org/pentaho/reporting/platform/plugin/cache/DeleteOldOnAccessCache.java
```diff
@@ -26,11 +26,7 @@
 import java.io.Serializable;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
tmorgner: Star imports are frowned upon by our coding standard. Please configure your IDE to avoid those.

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -282,6 +295,12 @@ public IReportContent getCachedContent( final String key ) {
     return cache.get( key );
   }
 
+  public Map<String, Serializable> getCachedMetaData( final String key ) {
```
tmorgner: THere is no reason to have this method as public, I suppose. No one can call this method from elsewhere.

test-src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutputTest.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+
+public class CachingPageableHTMLOutputTest {
```
tmorgner: I would argue that this is an integration test, as you are testing the whole stack of classes here. Name it with IT at the end.

src/org/pentaho/reporting/platform/plugin/cache/DeleteOldOnAccessCache.java
```diff
@@ -83,6 +79,22 @@ protected void setMillisToLive( final long millisToLive ) {
...
 
+  /**
+   * Cleans old entries Saves value with timestamp
+   *
+   * @param key   key
+   * @param value value
+   * @param metaData metaData
+   * @return success
+   */
+  @Override public boolean put( final String key, final IReportContent value,  Map<String, Serializable> metaData ) {
```
tmorgner: This method deserves to have an unit test.

src/org/pentaho/reporting/platform/plugin/cache/AbstractReportContentCache.java
```diff
@@ -41,10 +42,19 @@ public boolean put( final String key, final IReportContent value ) {
   }
 
   @Override
+  public boolean put( final String key, final IReportContent value, Map<String, Serializable> metaData ) {
```
tmorgner: If you reroute the other put to call this method with the new map as third parameter, you automatically get code coverage on here and less duplicate code.

dzmitry-hryhoryeu:
Thank you for review. Changes have been made according to your recommendations.

tmorgner:
There are test failures.

dzmitry-hryhoryeu:
All tests passed.

PR #379, opened 2016-06-21
BACKLOG-8067 - Add HTML email output type

EgorZhuk:
@tmorgner please take a look

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -1384,7 +1384,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
       },
 
       _isDownloadableFormat: function (mime) {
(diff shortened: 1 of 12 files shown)


EgorZhuk:
Junits for DefaultReportOutputHandlerFactory adeed.

tmorgner:
Can you check why there is a failing test?

EgorZhuk:
I will add more tests to finally resolve this floating coverage.

EgorZhuk:
@tmorgner  :tada: by the way did you know that we have 17,5k lines of production code in plugin and 14,5k lines of tests :+1:

tmorgner:
The code ratio is normal. With test setup and validation code (plus helpers) I would not be surprised to even see a 2:1 ratio of test to production code.

Its like the dreaded math-proofs: a^2 + b^2 = c^2 is easy to write, but proof its correctness to the teacher and you fill a sheet of paper. Thank you for bringing back bad memories  :)

PR #381, opened 2016-06-23
[BACKLOG-8335] Fix styling for the error message dialog in Onyx theme

Aliaksandr-Kastenka:
@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -21,8 +21,9 @@ define([
   "dijit/registry",
   "common-ui/prompting/api/PromptingAPI",
   "common-ui/jquery-clean",
-  "common-ui/underscore"
-], function(util, Messages, registry, PromptingAPI, $, _) {
+  "common-ui/underscore",
+  "cdf/dashboard/Utils"
+], function(util, Messages, registry, PromptingAPI, $, _, Utils) {
 
   var _api =  new PromptingAPI('promptPanel');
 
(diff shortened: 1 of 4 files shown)


package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -344,14 +344,18 @@ define([
       showMessageBox: function( message, dialogTitle, button1Text, button1Callback, button2Text, button2Callback, blocker ) {
         var messageBox = registry.byId('messageBox');
 
-        messageBox.setTitle(dialogTitle);
-        messageBox.setMessage(message);
+        messageBox.setTitle('<div class="pentaho-reportviewer-errordialogtitle">' + dialogTitle + '</div>');
+        messageBox.setMessage('<div class="pentaho-reportviewer-errordialogmessage">' + message + '</div>');
```
tmorgner: Are you sure this is correct? If you'd wanted to add a style class, I would expect that you set it in the template, not here as part of the string.

Aliaksandr-Kastenka: I want to make it similar to:

tmorgner: Woha. Now there is some seriously bad coding. But OK, if you want to use raw HTML, then at least do it safely. You must properly encode the contents of "dialogTitle" and "message" or you can easily create HTML and Script injections and other nasty surprises. 

You can use the escapeHTML method of "cdf/dashboard/Utils" for that purpose.

Aliaksandr-Kastenka:
@tmorgner 
Pull-request updated.

PR #382, opened 2016-06-23
[BACKLOG-9027] - Cenerated content of scheduled paginated report: mis…

dzmitry-hryhoryeu:
[BACKLOG-9027] - Cenerated content of scheduled paginated report: missing comma separator in page number field

@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/paginationTemplate.html
+++ b/package-res/reportviewer/paginationTemplate.html
@@ -49,6 +49,7 @@
       float: left;
       overflow: hidden;
       border: none;
+      width: inherit;
     }
 
     .pagePrev {
@@ -92,13 +93,14 @@
       border: 1px solid #b3b3b3;
       margin: 0 0.1em;
       float: left;
-      width: 2.1em;
+      width: 2.6em;
     }
 
     .pageNumberInputContainer {
       float: none !important;
       position: relative;
       overflow: hidden;
+      width: inherit;
     }
 
     .pageTotal {
@@ -127,7 +129,7 @@
             <div id="right" class="pageNext"></div>
             <div class="pageNumberInput">
               <div class="pageNumberInputContainer">
-                <input id="current" type="number"/>
+                <input id="current" type="text" pattern="[0-9]*"/>
               </div>
             </div>
             <div id="overall" class="pageTotal">/ 1</div>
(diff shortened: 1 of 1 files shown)


package-res/reportviewer/paginationTemplate.html
```diff
@@ -238,6 +238,10 @@
     );
   }
 
+  function numberWithCommas(x) {
+    return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+  }
```
tmorgner: If you want to format a number with thousands indicators, I would suggest you use locale aware code for that. 

For IE11 and similar modern browsers you can use "Number.prototype.toLocaleString()" (use the "useGroupingSeparator flag). If that method is not available (old browser etc.) you can fall back to your RegExp solution. But get it localized by default. 

Otherwise you will be surprised that not everyone uses comma for this purpose. Not everyone uses the US style system. (Given Russia and Belarus use comma as decimal separator lets not make everything conform to the English :) )

dzmitry-hryhoryeu: Amended.

package-res/reportviewer/paginationTemplate.html
```diff
@@ -238,6 +240,21 @@
     );
   }
 
+  function toLocaleStringSupportsLocales() {
+  var number = 0;
+  try {
+    number.toLocaleString('i');
+  } catch (e) {
+    return e.name === 'RangeError';
```
tmorgner: Put your old regexp-based solution into the catch block here. This way, if "toLocaleString" is not available, the user will get at least a result even though it is not in the correct locale.

package-res/reportviewer/paginationTemplate.html
```diff
@@ -238,6 +240,21 @@
...
+  }
+  return false;
```
tmorgner: Should that be a "true"? Your "getLocalizedValue" function implies as much here.

tmorgner:
Code style: You now have two functions and format the number twice (one to check for errors, second to return the formatted value. 

You could have simply used:

     try {
      return number.toLocaleString('i');
    } catch (e) {
       return return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
     }

here. This way, if native formatting is supported, you use that, else you fall back to the english format using your old style regexp code.

package-res/reportviewer/paginationTemplate.html
dzmitry-hryhoryeu: The bug was in that that in PUC there is a comma, but in this template there no comma. As far as I understand in PUC comma added according to locale, so if we will add a comma in paginationTemplate.html in case there is no support toLocaleString() method than we will get different formats in PUC and paginationTemplate.html. And this will lead to a new bug. If I understand correctly all.

tmorgner: This method here does not actually handle the case when the method "toLocaleString" does not exist. 

When the method itself is not supported, you wont get a "range-error", you get a different error depending on what browser you use. So checking for a specific error code is rather optimistic - especially when we are dealing with really really old browsers or email clients.

Play it safe and catch everything. 

And why test for the options-argument when your real method call never uses it? All you need here is to check whether the call to the method with no parameters will succceed. If you dont need the extended options, dont test for them.


As for the different behaviour between PUC and this template:

Locale handling in PUC is implemented via Dojo. As this is a standalone HTML file, with no server to provide libraries, we have to rely on whatever is built into the browser. The "toLocaleString" method is supported in all the latest browsers, but older browsers and email clients may not have that method. 

If a report is scheduled and has enough rows to trigger the auto-scheduling, the exporter will use this template to produce the report - so we better make sure that we are as compatible as possible. 

And yes, our behaviour will not be strictly the same as in the BI server if the user happens to use an old HTML viewer. But with the combination of "toLocaleString" method as default and the fallback to the hard-coded "en-US" style string format, we at least try our best and return a sensible fall-back behaviour. 



And as I said in the other comment: once you have formatted the number in the try block - why throw the result away? Use it.

tmorgner:
@dzmitry-hryhoryeu please investigate the build errors you see here. If you are sure the errors are not in your code, engage the build team as necessary. Commits that break that badly are not mergeable. (Same applies to your other pull request #378 )

dzmitry-hryhoryeu:
There are no problems except coverage.

PR #384, opened 2016-06-27
[BACKLOG-7980] - parameter dependencies server side discovery func.

prakapenka:
Provide better parameter handling to avoid multiple posts
to the server for revalidations.

Added new methods to discover cascading parameters as far as write this
info in responce xml.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
+++ b/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
@@ -27,6 +27,7 @@ import java.text.SimpleDateFormat;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


tmorgner:
@prakapenka There are check style and unit test failures. The unit test failure may go away once PR #387 is merged, so at that point please rebase and see if that helps. If not, then please investigate the cause of the error.

tmorgner:
@prakapenka  There are test failures.

prakapenka:
@tmorgner - *newly fixed tests != newly broken. Wingman shows no mercy.

tmorgner:
Sure enough. But a broken test here will break on CI as well. So please investigate and fix it. If you need guidance or hints on what may be the cause have a chat with @EgorZhuk - he has been very successful at hunting a lot of similar problems recently.

EgorZhuk:
@tmorgner the test is in Newly Fixed Tests section. And it is now fixed because Dmitry rebased on latest master with my fix :)

prakapenka:
nobody shows mercy at all.

tmorgner:
WTF. Why a frigging red mark. Gosh this is the second or third time that wingman message caught me offguard.

PR #387, opened 2016-06-28
CLEANUP - Fix naughty tests

EgorZhuk:
It seems that failing test showed a bug in executor. As far as I understand originally future callbacks were executed in the thread from wich they were added. From the guava code I found out that these callbacks lock the thread and simply wait for result. 

`addCallback(future, callback, MoreExecutors.sameThreadExecutor());`

We had a very bad situation when a bunch of long running scheduled reports could consume all threads from the tomcat thread pool and dramatically affect response time.

With the fix callbacks are now executed in our own thread pool.

@tmorgner please take a look.


pentaho-nbaker:
Awesome catch

tmorgner:
Nice catch. Thank god for unit tests :)

EgorZhuk:
@tmorgner I will rebase the branch

tmorgner:
Could you quickly you rebase the branch for a quick merge?

EgorZhuk:
@tmorgner done

PR #389, opened 2016-06-30
[BACKLOG-9031] - Different font of navigation bar in report viewer an…

dzmitry-hryhoryeu:
[BACKLOG-9031] - Different font of navigation bar in report viewer and generated content of scheduled paginated report.

@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/paginationTemplate.html
+++ b/package-res/reportviewer/paginationTemplate.html
@@ -1,13 +1,16 @@
+<!DOCTYPE html>
 <html>
 <head>
+  <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300,700' rel='stylesheet' type='text/css'>
   <style>
-    html, body {
+    html, body, input {
       margin: 0px;
       padding: 0px;
-      font-family: OpenSansRegular, sans-serif;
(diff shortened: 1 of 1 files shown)


package-res/reportviewer/paginationTemplate.html
```diff
@@ -1,13 +1,16 @@
...
 <html>
 <head>
+  <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300,700' rel='stylesheet' type='text/css'>
   <style>
-    html, body {
+    html, body, input {
       margin: 0px;
       padding: 0px;
-      font-family: OpenSansRegular, sans-serif;
+      font-family: 'Open Sans' !important;
```
tmorgner: open sans in itself is not enough. You should provide fallback fonts - at minimum the "sans-serif" must be in that list. This way any browser that does not have access to "Open Sans" (and many dont), then at least they show a sans-serif font instead of a browser specific font. 

So font-family should be

       'Open Sans', sans-serif

dzmitry-hryhoryeu: Exactly! You're right as always. Added.

PR #390, opened 2016-07-08
[BACKLOG-9357] - Message that row limit is reached is not shown in ca…

dzmitry-hryhoryeu:
[BACKLOG-9357] - Message that row limit is reached is not shown in case report is reduced to 1page

@tmorgner 
Please review.


src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -307,12 +309,25 @@ private synchronized void persistContent( final String key, final IReportContent
...
 
+  private Map<String, Serializable> updateQueryLimitReachedFlag( Map<String, Serializable> metaData ) {
+    if ( metaData == null ) {
+      metaData = new HashMap<>();
+    }
+    metaData.put( IS_QUERY_LIMIT_REACHED, processor.isQueryLimitReached() );
+
+    return metaData;
+  }
+
```
tmorgner: In case the "persistContent" method with 3 parameters is no longer used (and it seems it isn't), then please remove it. It is a private method, so it has no longer any reason to exist.

PR #392, opened 2016-07-11
[CLEANUP] - Remove unused method.

dzmitry-hryhoryeu:
[CLEANUP] - Remove unused method.

@tmorgner 
Please review.


tmorgner:
Lets ignore wingman's chatter about lower code coverage. Assuming no unit-tests changes (and here it clearly does not) removing coding brings up coverage, never lowers it.

PR #393, opened 2016-07-12
BACKLOG-7817 - Location prompting in reportviewer tab

EgorZhuk:
@tmorgner please review, junits are on the way

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -50,6 +50,13 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
       _previousPage: 0,
       _reportUrl : null,
       _handlerRegistration : null,
+      _locationPromptCancelHandlerRegistration: null,
+      _locationPromptOkHandlerRegistration: null,
+      _locationPromptFinishHandlerRegistration: null,
+      _locationPromptAttachHandlerRegistration: null,
+      _locationPromptFinished: null,
+      _locationOutputPath: null,
+      _manuallyScheduled: null,
 
       _bindPromptEvents: function() {
         var baseShowGlassPane = this.reportPrompt.showGlassPane.bind(this.reportPrompt);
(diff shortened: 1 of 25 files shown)


tmorgner:
@EgorZhuk Test failures make Wingman puke. :)

package-res/reportviewer/reportviewer.js
```diff
@@ -921,16 +938,92 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
 
+              registry.byId('feedbackScreen').hide();
               if (me.reportPrompt._promptForLocation) {
-                //Open location dialog here - see the example of location update below
-             /*   var urlUpdateLocation = url.substring(0, url.indexOf("/api/repos")) + '/plugin/reporting/api/jobs/' + me._currentReportUuid + "/schedule/location?folderId=5a8ba6eb-3d98-4a5c-9357-4a17848925a9&newName=hello_there";
-                pentahoPost(urlUpdateLocation);*/
-              }              
-              
+
+                if(window.top.mantle_addHandler) {
```
tmorgner: UpdateReportContentCore is already a huge monster of a function. Is there a chance to refactor this new block so that it is smaller and easier to read? Taking the block out into a new function would be a sweet start, and each of the functions look as if they can be standalone functions (and thus wont pollute the reading flow).

Even the json-parsing parts can be simplified by refactoring it (and the status check) into a function like this:

    (bool) 
    function isStatus (result, expectedStatus) {
      // happy path only, try catch just returns false then ...
      var status = JSON.Parse(result).status 
      var args = Array.prototype.slice.call(arguments, 1);
      return args.includes(expectedStatus);
    }

This cuts out a lot of the ugly clutter that JavaScript forces you to create for such a simple task.

package-res/reportviewer/reportviewer.js
```diff
@@ -1252,6 +1425,60 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
+        $("#glasspane").css("background", "");
+        $("#glasspane").css("display", "block");
+      },
+      
+      _removeLocationPromptHandlers: function(){
+        if (window.top.mantle_removeHandler) {
+          if(this._locationPromptCancelHandlerRegistration) {
+            window.top.mantle_removeHandler(this._locationPromptCancelHandlerRegistration);
+          }
+          if(this._locationPromptOkHandlerRegistration) {
```
tmorgner: Where is the removal of the "finish" handler? Three go in, two come out, and one parties on to create a memory leak :)

tmorgner:
Looks good to me. However, with all the back-and-forth to safely schedule and handle all the conflicts that might bring, I think we need to add a extra case "Provide documentation on the API interactions" to the backlog. I sure as hell will not remember any of the gory details in 6 month time. :)

tmorgner:
@EgorZhuk This PR has code style issues. Unused import and formatting.

#### Checkstyle violations

The code submitted is not complying with one or more of our guidelines. To perform these checks locally, [please use this tool](https://github.com/pentaho/pentaho-coding-standards).

###### /home/buildguy/project/src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncExecutor.java

> :heavy_exclamation_mark: Unused import - java.io.IOException. at Line: 38 Column: 8


###### /home/buildguy/project/test-src/org/pentaho/reporting/platform/plugin/async/PentahoAsyncReportExecutorTest.java

> :heavy_exclamation_mark: 'block' child have incorrect indentation level 5, expected level should be 6. at Line: 1,018

tmorgner:
One more checkstyle fix, please. Wingman demands it :)

EgorZhuk:
Need to work on tests stability.

tmorgner:
Aye. Loads of these classes are not tested directly, only indirectly via the integration tests or other higher-level tests. ReportContentImpl is a classic example of that. I think the various HTMLOutput implementations should be easy to cover as well, as they have a comparatively simple interface. 

I'd be more than happy if both ReportContentImpl and the (Abstract|Stream)HTMLOutput are explicitly covered. These would be the highest value targets - the test coverage on the other classes may fluctuate a bit, but as it is only by a few minor percent points, I think we can ignore them for now.

EgorZhuk:
At least we stabilized it :)

PR #395, opened 2016-07-18
BACKLOG-7075 - Set manual interruption flag on cancel

EgorZhuk:
@tmorgner please take a look 
Connected with https://github.com/pentaho/pentaho-reporting/pull/794


tmorgner:
Looks good so far. Lets wait for Wingman's approving nod, and then we merge.

EgorZhuk:
@tmorgner it depends on reporting-engine, the build will fail. I will work on reporting engine and then update this one.

pentaho-nbaker:
Try to avoid synchronized methods and use Atomics and Locks instead. For this it's fine, but just keep it in mind for the future.

EgorZhuk:
@pentaho-nbaker thank you, I will

PR #398, opened 2016-07-20
[BACKLOG-9157] - Dashboard: "Run in Background" button not appear bac…

dzmitry-hryhoryeu:
[BACKLOG-9157] - Dashboard: "Run in Background" button not appear back after changing edit mode to view.

Please, do not merge before pull https://github.com/pentaho/pentaho-platform-plugin-dashboards/pull/312 being merged.

@tmorgner 
Please review.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -57,6 +57,8 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
       _locationPromptFinished: null,
       _locationOutputPath: null,
       _manuallyScheduled: null,
+      _scheduleScreenBtnCallbacks: null,
+      _editModeToggledHandler: null,
 
       _bindPromptEvents: function() {
         var baseShowGlassPane = this.reportPrompt.showGlassPane.bind(this.reportPrompt);
@@ -70,6 +72,10 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         this.reportPrompt.api.event.afterUpdate(this.afterUpdateCallback.bind(this));
         this.reportPrompt.api.event.afterRender(this.view.afterRender.bind(this.view));
 
+        if (typeof parent.pho !== 'undefined' && typeof parent.pho.dashboards !== 'undefined') {
+          this._editModeToggledHandler = this.editModeToggledHandler.bind(this);
+          parent.pho.dashboards.addEditContentToggledListener(this._editModeToggledHandler);
+        }
       },
 
       load: function() {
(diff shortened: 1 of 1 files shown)


tmorgner:
@EgorZhuk , @dzmitry-hryhoryeu One more error in AsyncIT#testLogout. 
"java.lang.IllegalStateException: Cannot query state until job is added to the executor."

dzmitry-hryhoryeu:
@tmorgner 
Please review.

package-res/reportviewer/reportviewer.js
```diff
@@ -826,12 +830,36 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
         }else {
           scheduleScreenBtnCallbacks.shift();
+          dlg.hideBackgroundBtn();
         }
 
         dlg.callbacks = scheduleScreenBtnCallbacks;
         return dlg;
       },
 
+      renderBackgroundBtnListener: function (editMode) {
```
tmorgner: Just one change: "renderBackgroundBtnListener" is a horrible name for an event handler. functions starting with a verb normally signify an normal method, so here I would expect that something is rendered. 

Please name it: editModeToggledHandler - this makes it clear that it is an event handler, and what event it handled.

And change the name of "dlg2" to "feedbackScreen" (again, readability - this way you can look at the variable and see instantly what it is. "dlg2" is meaningless compared to "feedbackScreen".

package-res/reportviewer/reportviewer.js
```diff
@@ -826,12 +830,36 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
...
+        var dlg2 = registry.byId('feedbackScreen');
+        var scheduleScreenBtnCallbacks = _scheduleScreenBtnCallbacks;
+        if (dlg2) {
+          if (editMode) {
+            dlg2.hide();
+            scheduleScreenBtnCallbacks.shift();
+            dlg2.hideBackgroundBtn();
```
tmorgner: is it safe to modify this array in the "shift" operation here? What happens if you toggle between edit mode and view mode multiple times? 

Wouldnt it be safer to treat the array as immutable and to have 

    dlg2.callbacks = [ scheduleSchreenBtnCallbacks[1]]; 

here and 

     dlg2.callbacks = scheduleScreenBtnCallbacks; 

in the else branch?

package-res/reportviewer/reportviewer.js
```diff
@@ -1516,6 +1544,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         var me = this;
         //Update current values
         me._currentReportStatus = mainJobStatus.status;
+        _currentReportStatus = mainJobStatus.status;
```
tmorgner: If this line is needed, then you are accessing "_currentReportStatus" somewhere without qualifying it with the "this" or "me" reference. Fix your new "renderBackgroundBtn" function and qualify access with "this", and therefore access the correct member variable of your current object..

package-res/reportviewer/reportviewer.js
```diff
@@ -70,6 +71,9 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         this.reportPrompt.api.event.afterUpdate(this.afterUpdateCallback.bind(this));
         this.reportPrompt.api.event.afterRender(this.view.afterRender.bind(this.view));
 
+        if (typeof parent.pho !== 'undefined' && typeof parent.pho.dashboards !== 'undefined') {
+          parent.pho.dashboards.addEditContentToggledListener(this.editModeToggledHandler.bind(this));
```
tmorgner: one last thing: Every add should at some point have a remove, right? Otherwise, when the dashboard decides to remove your iframe, you will be left with a memory leak and/or an invalid function.

Have a look at how the mantle-handler does this job here, and do the same.

dzmitry-hryhoryeu:
Listener removed before the dashboard update iframe in https://github.com/pentaho/pentaho-platform-plugin-dashboards/pull/313

PR #400, opened 2016-07-22
[BACKLOG-7980] - Provide better parameter handling to avoid multiple …

Aliaksandr-Kastenka:
…posts to the server for revalidations
Goes with https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/809

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -36,6 +36,8 @@ define([
       _dialogThreshold: 1500,
       _promptForLocation: null,
       _isReportHtmlPagebleOutputFormat : null,
+      _oldParameterDefinition : null,
+      _oldParameterSet : null,
 
       /**
        * Gets the prompt api instance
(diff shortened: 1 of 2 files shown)


Aliaksandr-Kastenka:
@tmorgner 
Please review.

package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -73,10 +75,140 @@ define([
...
+          return false;
+        }
+        return true;
+      },
+
+      compareParameters: function(oldParameterSet, currentParameterSet) {
+        var changedParameters = [];
+
+        $.each(oldParameterSet, function (i, parametersGroup) {
+          if("parameters" == parametersGroup.group) {
```
tmorgner: Please make this test reverse: Test for "system" as group-name. This is a well-defined hard-coded name encapsulating all system level parameters like output-target, scheduling etc. This way you leave it open for other implementations to define additional parameter-groups. "parameters" is not one of the prescribed names, it is more or less just an implementation detail of the current level of support.

package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -73,10 +75,140 @@ define([
...
+        var currentParameterSet = this.api.operation.getParameterValues();
+
+        // compare currentParameterSet with oldParmaeterSet. Return an array of changed parameters. (More than one can change if we use the API).
+        var changedParameters = this.compareParameters(this._oldParameterSet, currentParameterSet);        
+        return changedParameters;
+      },
+
+      canSkipParameterChange: function(names) {
+        var outputTarget;
+        $.each(this._oldParameterSet, function (i, parametersGroup) {
```
tmorgner: This test is to restrictive. The parameter-UI in non-async mode will send the same parameter set twice, once for renderMode=XML and one for renderMode=PARAMETER. 

Implement a general rule that states that if any of the "system" parameters or the "output-target" changes, then you don't skip this request.

tmorgner:
Two more things: 

Your code gets unnecessarily complex by maintaining the group as extra level in your "oldparameterSet". 

It is guaranteed that the name of a parameter is unique across all groups (as parameter are sent as URL/form parameter). You only need that group-name while you compare the values, so a simplified map of a structure like this would solve your data-access needs in an easier way:

param-set = {
   paramName: { 
     value: null, 
     group: null
   }

   paramName2: { 
     value: null, 
     group: null
   }
}

This way you can directly lookup each parameter by name (without having to iterate through groups to then iterate through values). 


Second: Can you add some karma unit-tests for this code too? This should test that:

- system parameter changes always trigger a server call
- output target changes always trigger a server call
- the dependency-list is correctly interpreted:
  -- leaf parameters don't trigger a server call
  -- parameter marked as "always-validate" always trigger a server call
  -- changes in a parameter with dependent other parameters triggers a server call.
- when handling the update in the client: 
  -- changed values are correctly set in the parameter-set
  -- the limited parameter-validation (mandatory check from common-ui) happens
  -- error lists are updated (including setting new errors and clearing existing errors).

There is a set of specs in test-js and your methods are exposing a very clean interface and work on a very well-defined set of inputs. Therefore creating tests wont be overly hard, especially given that there are similar tests for that class (reportviewer-prompt) already.

tmorgner:
@Aliaksandr-Kastenka this pull request needs your attention.

Aliaksandr-Kastenka:
@tmorgner 
I am preparing update at the moment

Aliaksandr-Kastenka:
@tmorgner 
Pull-request updated.

> system parameter changes always trigger a server call

I believe "::session" parameter(generated by server) will be exceptional case, as for me, sometimes I got different "::session" with every new request.

> when handling the update in the client: -- changed values are correctly set in the parameter-set -- the limited parameter-validation (mandatory check from common-ui) happens -- error lists are updated 

This test already exists(in previous pull-request to common-ui project https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/809)

PR #401, opened 2016-07-25
BACKLOG-9160 - Pre-generate async jobs ids

EgorZhuk:
@tmorgner please take a look

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
+++ b/src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
@@ -25,6 +25,7 @@ import org.pentaho.platform.api.repository2.unified.RepositoryFile;
 import org.pentaho.platform.engine.core.audit.MessageTypes;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 import org.pentaho.reporting.engine.classic.core.modules.output.table.html.HtmlTableModule;
+import org.pentaho.reporting.platform.plugin.async.IJobIdGenerator;
 import org.pentaho.reporting.platform.plugin.async.IPentahoAsyncExecutor;
 import org.pentaho.reporting.platform.plugin.async.PentahoAsyncExecutor;
 import org.pentaho.reporting.platform.plugin.async.PentahoAsyncReportExecution;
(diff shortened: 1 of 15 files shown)


src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
```diff
@@ -87,23 +92,26 @@ public void createReportContent( final Serializable fileId, final String path,
...
       asyncExec = new PentahoAsyncReportExecution( path, reportComponent, handler, userSession, instanceId, audit );
 
-    IPentahoAsyncExecutor executor =
+    final IPentahoAsyncExecutor executor =
       PentahoSystem.get( PentahoAsyncExecutor.class, PentahoAsyncExecutor.BEAN_NAME, null );
     // delegation
     if ( reportComponent.validate() ) {
-      UUID uuid = executor.addTask( asyncExec, userSession );
+      final String precomputedId =
+        getRequestParameters().getStringParameter( PRECOMPUTED_ID, UUIDUtil.getUUIDAsString() );
```
tmorgner: UUIDUtil is the most useless class ever. Whats wrong with 

    java.util.UUID.randomUUID().toString();

tmorgner:
I slightly object to letting the untrusted client generate IDs. (JavaScript can be changed so easily, and forging requests is just as easy). Never trust the client.

Why not:

     Client: POST /reporting/api/job/reserveJob
     Server: { reserved-id: "1234-5678-..." }
     Client: POST /reporting/reportjob?reservedId=1234-5678-...
    (validate that the id requested has been previously reserved, otherwise discard the paramter-value and regenerate an new ID)

EgorZhuk:
@tmorgner I've updated the implementation

src/org/pentaho/reporting/platform/plugin/async/IJobIdGenerator.java
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright 2006 - 2016 Pentaho Corporation.  All rights reserved.
+ */
+
+package org.pentaho.reporting.platform.plugin.async;
+
+import org.pentaho.platform.api.engine.IPentahoSession;
+
+public interface IJobIdGenerator {
+
+  String generateId( IPentahoSession session );
```
tmorgner: Can this be strongly typed? We expect an UUID in the string, so let it be a UUID in the interface too. On the caller-site, you can easily parse the UUID via "UUID.fromString()" - which also validates the input. If the UUID given there is non-null, but invalid (ie illegal-arg exception is thrown), abort with an error. Aborting on invalid input is OK and safe and sane.

src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
```diff
@@ -113,22 +122,42 @@ public void createReportContent( final Serializable fileId, final String path,
     }
   }
 
+  private String getReservedId() {
```
tmorgner: This could easily be a UUID returning function.

src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
```diff
@@ -113,22 +122,42 @@ public void createReportContent( final Serializable fileId, final String path,
...
+    final IPentahoSession session = PentahoSessionHolder.getSession();
+    final IJobIdGenerator iJobIdGenerator = PentahoSystem.get( IJobIdGenerator.class );
+    final String failover = UUID.randomUUID().toString();
```
tmorgner: Make this an UUID....

src/org/pentaho/reporting/platform/plugin/BackgroundJobReportContentGenerator.java
```diff
@@ -113,22 +122,42 @@ public void createReportContent( final Serializable fileId, final String path,
...
+
+    UUID uuid = null;
+    try {
+      uuid = UUID.fromString( precomputedId );
+    } catch ( final IllegalArgumentException e ) {
+      logger.warn( "Wrong uuid came from client side: ", e );
+    }
+
+    if ( session != null && iJobIdGenerator != null && iJobIdGenerator.acquire( session, uuid ) ) {
+      return precomputedId;
```
tmorgner: Return "uuid" here ..

pamval:
@EgorZhuk this still shows some coverage drop in the classes you changed. Can you review please?

EgorZhuk:
@pamval sure, will work on it

tmorgner:
Given the broken Wingman code above, doing a manual review is the only option. 

Manual review looks good. Given that a locally running Jacoco contradicts the output of the junit-tests of the same run, I simply assume that Jacoco is broken and that the coverage metrics produced are random noise. Why else would it claim that "JobManagerTest#testReserveId" does not cover the "JobManager#reserveId" method. 

Merge then pray.

PR #404, opened 2016-07-29
[BACKLOG-9668] - Post-processing computation is not working

Aliaksandr-Kastenka:
All parameters can have formulas for post-processing and default-values, not only list parameters.


Aliaksandr-Kastenka:
@pamval 
Please review.

pamval:
@Aliaksandr-Kastenka unit test is failing, please take a look

PR #405, opened 2016-08-02
[BACKLOG-8197] - Apply custom ivy configuration to exlude dependency …

EgorZhuk:
…from the platform.


EgorZhuk:
The idea is simple - let's use custom ivy configuration to exlude conflicting dependency from the platform builds. EE platform already exludes it from libraries. Required classes are available on runtime for both ce and ee. 

@pamval @pentaho/orlandocalisbon

PR #406, opened 2016-08-03
[BACKLOG-9604] - Escape HTML to avoid encoding issues

EgorZhuk:
It increses the size of the resulting file but solves encoding issues.

@pentaho/orlandocalisbon @pamval please review


pamval:
@EgorZhuk why the drop in coverage?

EgorZhuk:
:computer: need to work on jax rs test stability

PR #410, opened 2016-08-11
[BACKLOG-9898] - Force styles inlining for HTML (E-Mail) output

EgorZhuk:
@tmorgner It works. The only concern I have - it can affect another reports running at the same time.


tmorgner:
Use the report's configuration instead of the global one. "MasterReport#getReportConfiguration()" will allow you to set it for just this one single instance.

EgorZhuk:
@tmorgner now using report configuration + junut

PR #411, opened 2016-08-12
[BACKLOG-9736] - Parameters are not resetted to default values on inv…

Aliaksandr-Kastenka:
…alid selections
Goes with https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/821

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -39,6 +39,7 @@ define([
       _isReportHtmlPagebleOutputFormat : null,
       _oldParameterDefinition : null,
       _oldParameterSet : null,
+      _defaultValuesMap : null,
 
       /**
        * Gets the prompt api instance
@@ -93,7 +94,7 @@ define([
               // add to changed
               changedParameters.push(parameter.name);
             }
-          } else {
+          } else if("" != parameter.value) {
             // add to changed
             changedParameters.push(parameter.name);
           }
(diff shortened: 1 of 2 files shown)


Aliaksandr-Kastenka:
For some reason Checkstyle violations repeated from previous commit despite the update.
@tmorgner 
Please review.

package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -93,7 +93,7 @@ define([
               // add to changed
               changedParameters.push(parameter.name);
             }
-          } else {
+          } else if("" != parameter.value) {
```
tmorgner: parameter.value can also legally be <null> in some cases. The server would produce an XML element "value" with null="true" as attribute. (see line 771 in ParameterXmlContentHandler). So unless the UI normalizes all occurences of <null> to "", I would extend the test here.

package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -222,6 +238,9 @@ define([
               paramDefn.autoSubmitUI = autoSubmit;
             }
 
+            if (this._oldParameterDefinition == null) {
+              this.extractDefaultValues(paramDefn);
+            }
```
tmorgner: I think that logic is an error waiting to happen. Consider the following flow:

(1) Intiial call: oldParamDef == null. ; current paramDefn updated to hold the extracted default values.
(2) Then a second, server validating update call to server: 

paramDefnCallback will be invoked from fetchParameterDefinition. 
this.oldParamDefn is no longer null, but paramDefn is a parsed object that has been produced from the XML. Your previously stored default-values were on a different paramDefn object, and are now lost. As "oldParameterDefinition" is not null, the "extractDefaultValues" is not called (and calling it now would give you the current selection, not sensible default values).

It might be wise to store the default-values separately from the paramDefn object. Personally I;d think there is nothing wrong just storing it as a simple map, keyed by parameter-name, here on the "reportviewer-prompt" (ie "this") instead.

package-res/reportviewer/reportviewer-prompt.js
Aliaksandr-Kastenka: At this point I am testing the following situation:
Comparing _oldParameterSet and currentParameterSet. 
currentParameterSet is a result of operation.getParameterValues() so it's not contains parameters with empty values.
_oldParameterSet contains empty parameters, so here I'm checking:
if currentParameterSet has such parameter AND _oldParameterSet has not such parameter AND parameter values is not empty - push it to the changedParameters.

package-res/reportviewer/reportviewer-prompt.js
Aliaksandr-Kastenka: @tmorgner 
My bad, this scenario slipped under my radar. Pull-requests updated.

PR #412, opened 2016-08-16
[BACKLOG-9595] - Clear cache from the reportviewer navigation panel

EgorZhuk:
@tmorgner please review

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
+++ b/src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
@@ -1,0 +1,56 @@
+/*
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License, version 2 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/gpl-2.0.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ *
+ * Copyright 2006 - 2016 Pentaho Corporation.  All rights reserved.
+ */
+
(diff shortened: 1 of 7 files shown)


EgorZhuk:
Wow, at least build is working, I will check the test

src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
```diff
@@ -0,0 +1,53 @@
...
+  @POST @Path( "clear" )
+  public Response clear() {
+    try {
+      final IPluginCacheManager iPluginCacheManager = PentahoSystem.get( IPluginCacheManager.class );
+      final IReportContentCache cache = iPluginCacheManager.getCache();
+      cache.cleanup();
+
+      final ICacheManager cacheManager = PentahoSystem.get( ICacheManager.class );
+
+      cacheManager.clearRegionCache( "report-dataset-cache" );
```
tmorgner: Can you add a call to "DataCacheFactory.getCache().getCacheManager().clearAll()" as well? In the default configuration it will clear the same cache as the "report-dataset-cache", but if an OEM replaces the DataCache-implementation, we will cover their implementation too.

test-src/org/pentaho/reporting/platform/plugin/async/AsyncIT.java
```diff
@@ -767,44 +767,6 @@ public void testScheduleNoSession() throws Exception {
     assertTrue( flag[ 0 ] );
   }
 
-  @Test
```
tmorgner: Yeah, its so nice to have randomized test results. Keeps everyone on their toes. :)

This test wont be missed as it is more busy testing the framework than testing the actual code, even though the statistics now show a gaping hole here. Can you make these validations a plain unit-test (ie test direct invocations of the JobManager) to see that recalculate and schedule have been called correctly?

EgorZhuk:
@tmorgner I've just moved the test to the separate class, didn't remove it :)

EgorZhuk:
@tmorgner hmm, do you know if wingman runs tests the same way as jenkins, I mean, does it use fork mode for each test class?

tmorgner:
I have no clue. Theoretically the fork-mode settings should have no impact at all as the same code is run (just that the setup code may be run more or less times depending on the fork-mode).

When I am trying to manually validate pull request coverage levels (and failing at it), I get wild and implausible results from the coverage tool - like a test for specific method being run successfully, but the coverage tool claims the method has never been called. Personally I treat the numbers above as the result of reading tea-leaves, and not as a scientific measure.

PR #413, opened 2016-08-17
[BACKLOG-10064] - Rotated text as images in emails

EgorZhuk:
@tmorgner please review


EgorZhuk:
Ok, let's get rid of jax-rs test and call JobManager itself.

tmorgner:
This pull request needs a rebase to resolve the conflict. And yeah, lets pray that the latest fixes in the tests make it more stable. I have a unconfirmable nagging feeling that Jacoco does not like multithreading or so.

EgorZhuk:
@tmorgner not so bad, as I imagined

EgorZhuk:
@tmorgner is it good enough or extra work on coverage is needed?

PR #414, opened 2016-08-17
[BACKLOG-10020] - Clear cache action

EgorZhuk:
@tmorgner please review


EgorZhuk:
Connected with https://github.com/pentaho/pentaho-platform/pull/3033

EgorZhuk:
@tmorgner pull request is updated. Actually null can came to the method if not in design mode and we need to handle it properly - to skip the rotation.

tmorgner:
Somehow Wingman does not like your style ;) 
Also: Can you add a simple unit test for it so that we maintain our level of Wingman happiness?

EgorZhuk:
@tmorgner the PR is updated

PR #415, opened 2016-08-18
[BACKLOG-10063] - Support pageable HTML in platform scheduler

EgorZhuk:
@tmorgner please take a look


src/org/pentaho/reporting/platform/plugin/output/PageableHTMLOutput.java
```diff
@@ -56,6 +56,9 @@
 import java.util.Map;
 
 public class PageableHTMLOutput implements ReportOutputHandler {
+
+  private String FORCE_ALL_PAGES = "org.pentaho.reporting.platform.plugin.ForceAllPages";
```
tmorgner: Make it "public static final" and put it in a central place. The "PentahoPlatformModule" class is a good place and copies what we do in the other reporting-modules. Then you can remove the duplicate string in the "SimpleReportingAction" - both can reference the central one.

tmorgner:
Looks sweet.

PR #417, opened 2016-08-22
[BACKLOG-10041] - change GET to POST not to overload url with parameters

prakapenka:
Lucky pentahoGet and penahoPost have almost same signature, so to remove parameters from url seems only change all pentahoGet('reportjob' ... ) to a post one implementation.


tmorgner:
This fix is almost boring ...

PR #420, opened 2016-08-24
[BACKLOG-6859] - As a report viewer, I want to be able to set a row l…

dzmitry-hryhoryeu:
[BACKLOG-6859] - As a report viewer, I want to be able to set a row limit on a report similar to PIR.

### DIFF
--- a/package-res/settings.xml
+++ b/package-res/settings.xml
@@ -12,5 +12,10 @@
     <auto-submit>false</auto-submit>
     <auto-submit-default>false</auto-submit-default>
 	-->
+
+    <!-- Enabled or disabled row limit control on UI. No value means disabled.-->
+    <query-limit-ui-enabled>false</query-limit-ui-enabled>
+    <!-- The maximum number of rows that will be rendered in a report viewer. -->
+    <query-limit>100000</query-limit>
     
 </settings>
--- a/src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
+++ b/src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
@@ -17,19 +17,7 @@
 
 package org.pentaho.reporting.platform.plugin;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.Serializable;
-import java.lang.reflect.Array;
-import java.net.URL;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
-import javax.print.DocFlavor;
-import javax.print.PrintService;
-import javax.print.PrintServiceLookup;
-
+import org.apache.commons.lang.math.NumberUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.pentaho.platform.api.engine.IAcceptsRuntimeInputs;
@@ -67,6 +55,19 @@ import org.pentaho.reporting.platform.plugin.output.FastExportReportOutputHandle
 import org.pentaho.reporting.platform.plugin.output.ReportOutputHandler;
 import org.pentaho.reporting.platform.plugin.output.ReportOutputHandlerFactory;
 import org.pentaho.reporting.platform.plugin.output.ReportOutputHandlerSelector;
+
+import javax.print.DocFlavor;
+import javax.print.PrintService;
+import javax.print.PrintServiceLookup;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
+import java.lang.reflect.Array;
+import java.net.URL;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
 
 public class SimpleReportingComponent implements IStreamingPojo, IAcceptsRuntimeInputs {
 
@@ -451,6 +452,15 @@ public class SimpleReportingComponent implements IStreamingPojo, IAcceptsRuntime
     }
     if ( inputs.containsKey( DASHBOARD_MODE ) ) {
       dashboardMode = "true".equals( String.valueOf( inputs.get( DASHBOARD_MODE ) ) ); //$NON-NLS-1$
+    }
+    if ( inputs.containsKey( ParameterXmlContentHandler.SYS_PARAM_QUERY_LIMIT ) ) {
+      try {
+        if ( inputs.get( ParameterXmlContentHandler.SYS_PARAM_QUERY_LIMIT ) != null ) {
+          this.getReport().setQueryLimit( checkAndGetUserInputQueryLimit( inputs.get( ParameterXmlContentHandler.SYS_PARAM_QUERY_LIMIT ), this.getReport().getQueryLimit() ) );
+        }
+      } catch ( Exception e ) {
+        log.warn( e.getMessage(), e );
+      }
     }
   }
 
@@ -1009,7 +1019,7 @@ public class SimpleReportingComponent implements IStreamingPojo, IAcceptsRuntime
         log.warn( Messages.getInstance().getString( "ReportPlugin.warnUnprocessableRequest", outputType ) );
         return 0;
       }
-      synchronized( reportOutputHandler.getReportLock() ) {
+      synchronized ( reportOutputHandler.getReportLock() ) {
         try {
           return reportOutputHandler.paginate( report, getYieldRate() );
         } finally {
@@ -1063,6 +1073,26 @@ public class SimpleReportingComponent implements IStreamingPojo, IAcceptsRuntime
         input = defaultValue;
       }
       return idx.cast( input );
+    }
+  }
+
+  protected int checkAndGetUserInputQueryLimit( Object userInputQueryLimit, int reportQueryLimit ) {
+    int systemQueryLimit = -1;
+
+    IPluginManager pm = PentahoSystem.get( IPluginManager.class );
+    if ( pm != null ) {
+      systemQueryLimit = NumberUtils.toInt( (String) pm.getPluginSetting( "reporting", "settings/query-limit", "-1" ), -1 );
+    }
+
+    int userQueryLimit = NumberUtils.toInt( (String) userInputQueryLimit, -1 );
+
+    if ( userQueryLimit > 0 ) {
+      return systemQueryLimit > 0 ? Math.min( userQueryLimit, systemQueryLimit ) : userQueryLimit;
+    } else {
+      if ( systemQueryLimit > 0 ) {
+        return reportQueryLimit > 0 ? Math.min( reportQueryLimit, systemQueryLimit ) : systemQueryLimit;
+      }
+      return -1;
     }
   }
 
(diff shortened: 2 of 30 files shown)


package-res/settings.xml
```diff
@@ -12,5 +12,13 @@
     <auto-submit>false</auto-submit>
     <auto-submit-default>false</auto-submit-default>
 	-->
+
+    <!-- The maximum number of rows that will be rendered in a report viewer. A zero value means no limit. -->
+    <maximum-query-limit>100000</maximum-query-limit>
```
tmorgner: The row-limits should be disabled by default so that we are compatible with the old versions. Admins can enable row-limits if they want to. 

A user-limit of 500 sounds unreasonable low and now what the case described. The case requests that the row-limit attribute of the report is used for a default value instead.

src/org/pentaho/reporting/platform/plugin/output/StreamHtmlOutput.java
```diff
@@ -62,6 +63,9 @@ public int generate( final MasterReport report,
     }
     try {
       sp.processReport();
+      if ( listener != null && listener instanceof IAsyncReportListener ) {
```
tmorgner: For this an all code like this: ReportListenerThreadHolder returns an IAsyncReportListener, so this check can be simplified by making the listener variable to be of type IAsyncReportListener .

src/org/pentaho/reporting/platform/plugin/SimpleReportingAction.java
```diff
@@ -880,8 +880,10 @@ public boolean _execute() throws Exception {
       }
     }
     // SP-307, BISERVER-9688
-
     final MasterReport report = getReport();
+    // Set unlimited queryLimit for reports running in "Run in background mode" in PUC.
+    report.setQueryLimit( -1 );
```
tmorgner: Don't delete the query limit. A user can set this limit on the master-report and would expect that limit to be honoured. So a blanked "make this report unlimited" operation like this is wrong. The report will be a freshly loaded report anyway, so actually: Dont do anything. Delete this row and the report will behave as it should.

src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -452,6 +453,15 @@ public void setInputs( final Map<String, Object> inputs ) {
     if ( inputs.containsKey( DASHBOARD_MODE ) ) {
       dashboardMode = "true".equals( String.valueOf( inputs.get( DASHBOARD_MODE ) ) ); //$NON-NLS-1$
     }
+    if ( inputs.containsKey( QUERY_LIMIT ) ) {
+      try {
+        if ( this.getReport() != null && inputs.get( QUERY_LIMIT ) != null ) {
```
tmorgner: getReport() never returns null. If no report-object is set, it will try to parse it from any of the other properties, and if that fails, it will throw an ResourceException.

tmorgner:
The logic seems sound and sane, so just a bit of cleanup of the issues I flagged up and it is ready to be merged.

tmorgner:
@dzmitry-hryhoryeu Ping. This pull request needs your attention.

src/org/pentaho/reporting/platform/plugin/SimpleReportingAction.java
dzmitry-hryhoryeu: You wrote in the letter that "•	A report that is run via the “Run in background/Schedule” option in PUC’s file browser never applies the row-limit". This code addresses this case. If we will not reset the limit, then the report will be generated with limit that user set in RD.

tmorgner: Never *sets* a row-limit. But that does not mean that we remove a row-limit that has been set by the user previously. This attribute is editable in the report-designer and therefore a user can set their own limits here. 

So in here you leave the query limit alone - if the user has defined (in PRD) that their report shall never return more than 10 rows, then this is what the user shall get.

src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -452,6 +453,15 @@ public void setInputs( final Map<String, Object> inputs ) {
...
+          this.getReport().setQueryLimit( Integer.parseInt( (String) inputs.get( QUERY_LIMIT ) ) );
```
tmorgner: And here: Check the existing query limit. There may be one on the report already. 

Also: The input here comes from the URL, so it can be any value. You will have to add additional check to ensure that the value you receive from the inputs is within the system- and user-limits set in the configuration. So the value you should set here is the minimum of either of these values (accounting for the meaning of <= 0 as "no limit", of course).

dzmitry-hryhoryeu: User cant't enter query limit more than system query limit - this check we have on UI. Why we need check existing query limit? If user enter 200 rows, we should set query limit 200 in report.

tmorgner: A Web UI is by definition always an untrusted source of input. You know that you can change URL parameters with relative ease and that you can even use other tools to send request to a server, right? I dont care what the UI code says, as long as I can send a request via HTTP I can program my own UI that ignores all *your* UI code. And then your limits, if not enforced on the server when the request is processed, are all for nothing.

At no point ever must you accept values from the outside world and NOT check them.

dzmitry-hryhoryeu:
@tmorgner please review

package-res/settings.xml
```diff
@@ -12,5 +12,10 @@
     <auto-submit>false</auto-submit>
     <auto-submit-default>false</auto-submit-default>
 	-->
+
+    <!-- Enabled or disabled row limit control on UI. No value means disabled.-->
+    <query-limit-enabled>false</query-limit-enabled>
+    <!-- The maximum number of rows that will be rendered in a report viewer. -->
+    <query-limit>100000</query-limit>
```
tmorgner: This is a bit misleading. We have at least two limits here:

(a) System-limit - the hard limit set by the administrator. No one can ever circumvent that. (Oh, and to be backward compatible, it has be *disabled* by default.)
(b) user-limit - the a default value used for the UI. Can be lower than the system-limit. This value would be used if the report itself does not define a value. 

But whether we should use the report's defined query-limit at all as a default value is another topic for tomorrows meeting. 

I would also rename the "query-limit-enabled" flag to "query-limit-ui-enabled" to make it clear that this flag controls the UI and user-limit, but does not enable/disable the system-limit.

src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -103,6 +104,7 @@
   public static final String DASHBOARD_MODE = "dashboard-mode"; //$NON-NLS-1$
   private static final String MIME_GENERIC_FALLBACK = "application/octet-stream"; //$NON-NLS-1$
   public static final String PNG_EXPORT_TYPE = "pageable/X-AWT-Graphics;image-type=png";
+  public static final String QUERY_LIMIT = "query-limit"; //$NON-NLS-1$
```
tmorgner: Kill this duplicate string, use ParameterXmlContentHandler.SYS_PARAM_QUERY_LIMIT in its place. This way there is only one source of truth.

tmorgner:
When you extract the total-rows from the report, you are casting to an concrete class (AsyncReportStatusListener) instead of working with the interface. 

Never cast to a concrete implementation. That is bad style. If your interface needs a new method, then add that method to the interface. 

Extend the interface to have the setTotalRows(..)  if you absolutely need that setter. But before you start:

Why *do* you need to set this value manually? 

For all non-caching outputs: If you get your content from the report-processor (not cached), you will receive a report-status-event that has a valid total-rows value for you. 

For CachingPageableHTMLOutput, you have two cases:
(1) Nothing valid is in the cache, you are regenerating content (method "regenerateCache(..)"). At that point, your IAsyncReportListener will be filled with the correct total-rows from the last event it received. You need a public getter for the existing "totalRows" property. But there is no need for a setter, as you get that from the report-processor already. 

You can then store the total-rows as metadata along with the content in the cache. 

(2) There is content in the cache. 
When you restore the cached content (line 197/199), you can easily ask the cache for the metadata of your content to retrieve the stored totalRows value, or you can alter the "getCachedContent( key )" call to return both the content and the metadata/rowCount. 

When you do restore your cached content, you already fire the report-progress-event to set all the values, so instead of asking the listener for the row-count, use the totalRows that you retrieved from the metadata. 


And last but not least: Your change to setQueryLimitReachedToListener(..) where you set the totalRows is not doing anything. Either you have a cached page and update the listener (including the totalRows) from the report-event (line 197,199) or you regenerate the cached page and get a report-progress-event during the regenerateCache(..) call. So whatever value you have set in "setQueryLimitReachedToListener" is guaranteed to be replaced by one of these two cases.

package-res/settings.xml
dzmitry-hryhoryeu: Yes, in the current implementation, we use defined query-limit as a default value.

dzmitry-hryhoryeu:
@tmorgner please review

src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -1066,4 +1076,24 @@ public String getJcrOutputPath() {
     }
   }
 
+  protected int checkAndGetUserInputQueryLimit( Object userInputQueryLimit, int reportQueryLimit ) {
+    int systemQueryLimit = -1;
```
tmorgner: If a report-query-limit is defined, you MUST use this value. You *may* limit it by the system limit (pending today's meeting results). If there is a reportQueryLimit, the userQueryLimit must NOT be used.

Please do me a favour and add a unit-test for this method where you test out the various combinations. Remember: New code needs testing anyway, and with proper tests you will detect a few logic flaws here.

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -241,7 +242,7 @@ IReportContent regenerateCache( final MasterReport report, final int yieldRate,
     throws ReportProcessingException {
     logger.warn( "Regenerating report data for " + key );
     final IReportContent result = produceCacheablePages( report, yieldRate, key, acceptedPage );
-    persistContent( key, result );
+    persistContent( key, result, null );
```
tmorgner: This will backfire. You are storing content without a totalRows indicator, so the next time you retrieve it from the cache, you are missing that information. 

The "produceCachablePages" method will update the listener state from the events that get fired during the report processing, but without persisting that information, you are producing an invalid state.

Create a unit-test where you can replay that scenario (retrieve from cache, but not using the "listener.isFirstPageMode") and check what your cache contains. You will find that you have no totalRows stored and that subsequent cache-retrievals will therefore not set the correct totalRows counter.

dzmitry-hryhoryeu: Thanks for review, you're absolutely right. Amended.

tmorgner:
Please add unit tests. It helps to prove that your code is working correctly.

tmorgner:
@dzmitry-hryhoryeu Still needs unit test coverage for newly added code.

tmorgner:
@dzmitry-hryhoryeu , @EgorZhuk Looks like this pull request is abandoned.

EgorZhuk:
@tmorgner yeep, I will close it, I will open a new one when the feature will be ready

PR #421, opened 2016-08-25
[BACKLOG-8744] - Report Viewer: Glass Pane doesn't disable navigation…

Aliaksandr-Kastenka:
… using keyboard allowing a report to be executed.

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -934,6 +934,8 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         // that the report content is being updated.
         me.reportPrompt.showGlassPane();
 
+        me._setParametersDisablement(true);
+
         var options = me._buildReportContentOptions();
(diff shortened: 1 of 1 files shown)


Aliaksandr-Kastenka:
@EgorZhuk @tmorgner 
Please review.

package-res/reportviewer/reportviewer.js
```diff
@@ -934,6 +934,8 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
         // that the report content is being updated.
         me.reportPrompt.showGlassPane();
 
+        me._setParametersDisablement(true);
```
tmorgner: Can you rename that function to "updateParametersDisabledState(..)" 

you are not setting a single property but manipulate a whole bunch of elements, so "set" is a misleading prefix. And I would argue that Disablement is not quite a suitable description for what you are doing (updating the disabled state).

EgorZhuk:
@tmorgner   @Aliaksandr-Kastenka is on vocation, please close this pull request. I've opened #423 with renamed function

PR #422, opened 2016-08-26
[BACKLOG-10292]-Fix refresh button styles

EgorZhuk:
@tmorgner please merge


EgorZhuk:
Why something is fixed? When it was broken? Seems wingman is insane.

tmorgner:
Dont worry, as long as there are no Java files involved, just ignore Wingman's noise. But I do love it when checking in a CSS file increases code coverage :)

PR #424, opened 2016-08-26
[backlog-8362] - enchase caclulation of frame scrollWidth/Height

prakapenka:
Double check for better consistency.


prakapenka:
^^^ one more small css fix in addition

prakapenka:
@tmorgner pls review

PR #426, opened 2016-08-31
[CLEAN-UP] - fix style and calculate frame size only once.

prakapenka:


prakapenka:
@tmorgner please review

PR #428, opened 2016-09-05
[BACKLOG-8607]-Send minimized parameters response from server

EgorZhuk:
Depends on https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/828
@tmorgner please review

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -101,7 +101,7 @@ define([
         });
 
         for (var parameter in currentParameterSet) {
-          if (!oldParameterSet.hasOwnProperty(parameter)) {
+          if (oldParameterSet && !oldParameterSet.hasOwnProperty(parameter)) {
             changedParameters.push(parameter);
           }
         }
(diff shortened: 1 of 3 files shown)


package-res/reportviewer/reportviewer-prompt.js
```diff
@@ -186,7 +186,16 @@ define([
       updateParameterDefinitionWithNewValues: function(callback, names) {
         var paramDefn = $.extend(true, {}, this._oldParameterDefinition); // clone previous(old) paramDefn
         for(var i=0; i<names.length; i++) {
-          this.api.util.validateSingleParameter(paramDefn, names[i], this.api.operation.getParameterValues()[names[i]], this._defaultValuesMap);
+          /*As far as we don't recieve all parameters from server we can loose
+            calculated formula results. Let's handle this case. */
+          var untrusted = this.api.operation.getParameterValues()[names[i]];
+          if(undefined === untrusted ){
+            var p = paramDefn.getParameter(names[i]);
+            if(p && p.attributes && p.attributes['post-processor-formula']){
```
tmorgner: Instead of latching on to the post-processor formula, you should be able to simply test for the "mandatory-server-validation" (or whatever its name) attribute/flag instead. This flag will be set to true if there is any chance that the value returned is not a fixed result (including post-processor formula's potential changes).

tmorgner:
Code looks good so far (apart from that you can use the "must-validate-on-server" flag instead of looking for the formulas directly). However, you closed the commons-ui pull-request and this code relies on it. 

Once there is a valid PR for commons-ui, I can happily merge this code.

tmorgner:
On a parallel note: AsyncIT.testRecalcFinished either still seems to yield random results or wingman's pre-merge build is pulling something really old. I start to hate the sight of that particular unit test :)

EgorZhuk:
Will work on test.

tmorgner:
@EgorZhuk Given that the test is a long running "fun" story lets move that to a separate PR when its ready.

PR #430, opened 2016-09-19
[BACKLOG-10689] - Report could be executed with not valid parameters.

dzmitry-hryhoryeu:
[BACKLOG-10689] - Report could be executed with not valid parameters.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
+++ b/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
@@ -723,7 +723,8 @@ public class ParameterXmlContentHandler {
 
             parameterElement.appendChild( attributeElement );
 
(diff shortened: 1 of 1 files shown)


tmorgner:
@dzmitry-hryhoryeu loads of broken tests need your attention.

EgorZhuk:
@tmorgner @dzmitry-hryhoryeu  seems the API has changed. Plsease update PentahoReportEnvironment with following changes:

1. Change Authentication package to  forg.springframework.security.core.Authentication 
2. Get an array from the collection returned by the new API authorities.toArray( new GrantedAuthority[ authorities.size() ] );

pedrofvteixeira:
@EgorZhuk @tmorgner @dzmitry-hryhoryeu please hold off on processing this one until I re-trigger a new wingman review

dzmitry-hryhoryeu:
@tmorgner please review

tmorgner:
For every parameter that has a post-processing formula, the server should have reported the parameter as "must-validate-on-server" = true. Fix that so that the server returns the correct state for the parameter, instead of adding a manual workaround-check on the client side.

PR #431, opened 2016-09-20
[BACKLOG-10673] - Report is not generated if deselecting simple param…

dzmitry-hryhoryeu:
[BACKLOG-10673] - Report is not generated if deselecting simple parameters with no dependencies.


dzmitry-hryhoryeu:
@tmorgner please review

PR #432, opened 2016-09-22
[BACKLOG-6859]-Row limit in reportviewer

EgorZhuk:
Lets' see wingman's report, some tests are definitely needed.

@tmorgner this one contains the work from #420 but UI is moved to common-ui + some scheduling logic  is implemented


EgorZhuk:
Not so long list :)

src/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -419,4 +433,16 @@ private CacheKeyException( final Throwable cause ) {
...
+
+  private int getReportTotalRows( final String key ) {
+    final Map<String, Serializable> metaData = getCachedMetaData( key );
+    if ( metaData != null ) {
+      final Object reportTotalRows = metaData.get( REPORT_ROWS );
+      if ( reportTotalRows != null ) {
+        return (int) reportTotalRows;
+      }
+    }
+    return 0;
```
tmorgner: If there is no valid result, I feel it might be not so nice to pretend you have one. Zero rows is a valid result if your table-model returned has no rows (SELECT * FROM x WHERE FALSE). So returning "-1" here might help keep the API's response clean.

If you return "0" here, you actually set the row-limit in the UI to zero (see RowLimitControl, line 110). If you return -1, you have a way of detecting this kind of bad state and you can leave the UI unchanged; and hopefully with the next full generation you get a valid result.

(You would only get to this point if the cache go cleared out between your first call to get the report-content and the second call here where you get the metadata. Would not happen in our implementation, but as the interfaces are public for OEMs, they can create their own backends which then may evict content at random points in time. The only way to avoid that race condition would be to combine the reading of the content and the metadata into one single operation so that either both items are available or none is. But lets not start with such a rewrite, lets handle the error in a predictable way instead.)

src/org/pentaho/reporting/platform/plugin/output/FastXLSOutput.java
```diff
@@ -59,4 +76,41 @@ public boolean supportsPagination() {
   public Object getReportLock() {
     return this;
   }
+
+  public static void createXLS( final MasterReport report, final OutputStream outputStream,
```
tmorgner: THere is a good case for removing that code here. Make the FastXLSOutput a subclasss of the normal XLSOutput and simply call "super.generate(..)" when you want to switch to the normal processing. 

Less code here means you have less new lines that Wingman complains about because of missing unit-tests :)

EgorZhuk:
@tmorgner I've applied the changes according to your comments. Some more work is needed to be done  on testing.

Unfortuantely jacoco doesn't take into account Powermock tests [issue on GitHub](https://github.com/jacoco/eclemma/issues/15)
But changes in 
org.pentaho.reporting.platform.plugin.async.PentahoAsyncReportExecution
org.pentaho.reporting.platform.plugin.JobManager.ExecutionContext
are covered by
org.pentaho.reporting.platform.plugin.async.ExecutionRestartTest
org.pentaho.reporting.platform.plugin.JobManagerContextTest

EgorZhuk:
Jacoco works better with powermock 1.6.4, updated the version.

src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -1066,4 +1076,24 @@ public String getJcrOutputPath() {
     }
   }
 
+  protected int checkAndGetUserInputQueryLimit( final Object userInputQueryLimit, final int reportQueryLimit ) {
```
tmorgner: Just one thing: 
If - on the URL - I pass in a user-limit that is higher than the limit that is already hard-coded on the report, this code will not give precedence to the hard-coded limit. So if I have a top-10 report, I can happily extend it to a Top-1000 report, and thus see data to which I am not supposed to have access. This is a great way to download a list of all customers, or do other evil things.

EgorZhuk:
Much more clear now, powermock is taken into account. Will work on the coverage.

EgorZhuk:
@tmorgner is it good enough?

src/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -1066,4 +1079,27 @@ public String getJcrOutputPath() {
     }
   }
 
+  int checkAndGetUserInputQueryLimit( final Object userInputQueryLimit, final int reportQueryLimit ) {
```
tmorgner: You are trusting the user's input. Lovely. What happens if I manipulate the URL to get a really large (larger than system limit) user limit into this method?

You'll need a Min(sysLimit, userLimit) check. 
Your unit-test misses the set of tests where the user limit is larger than the system limit while no report-limit is set.

The logic is as follows:

	# If the report has a defined row limit, ignore the user's input, but apply the system limit
	if haveReportLimit
	   if haveSystemLimit
	     return Min(systemLimit, ReportLimit);
	   else
	     return reportLimit
	else
           # system and/or user limit may be undefined. so we need a hell of a if-then-else block ..
	   if haveSystemLimit && haveUserLimit
	     return Min(userLimit, systemLimit);
	   else if haveSystemLimit
	     return systemLimit
	   else if haveUserLimit
	     return userLimit
	   else 
	     return -1

EgorZhuk:
@tmorgner updated the code and test

EgorZhuk:
The more wingman runs - the more confidence in the tests stability :)

EgorZhuk:
@tmorgner pull request updated

EgorZhuk:
@tmorgner should I give some more ❤️ to this PR?

tmorgner:
Lets merge ..

tmorgner:
@EgorZhuk bring the changes back in with a new pullrequest either once the PIR changes are ready or remove the abstract function that is holding up the QA build - then we can merge earlier and without risk. A dummy default implementation on the interface should do that trick.

PR #434, opened 2016-09-26
[BACKLOG-8744] Report Viewer: Glass Pane doesn't disable navigation u…

Aliaksandr-Kastenka:
…sing keyboard allowing a report to be executed(additional pull-request).

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -1137,6 +1137,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
                       _Messages.getString('ErrorPromptTitle'));
                   registry.byId('feedbackScreen').hide();
                   me._hideAsyncScreens();
+                  me._updateParametersDisabledState(false);
 
                   logger && logger.log("ERROR: Request status - FAILED");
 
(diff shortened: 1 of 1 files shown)


Aliaksandr-Kastenka:
@tmorgner 
Please review.

tmorgner:
@EgorZhuk Seems another Async test plays up when run on Wingman. (PentahoAsyncReportExecutorTest). As this is a JavaScript only change, there is no way the new code can have caused the failure.

EgorZhuk:
@tmorgner seems https://github.com/pentaho/pentaho-platform-plugin-reporting/blob/master/src/org/pentaho/reporting/platform/plugin/async/AbstractAsyncReportExecution.java#L166 is the cause. The method is synchronized but it doesn't save us from race condition because listener has its own sychroniztion.  Cahging the code to provided below fixes the issue.

```
public synchronized boolean schedule() {
    synchronized( listener ){
      if ( listener.isScheduled() ) {
        return false;
      } else {
        listener.setStatus( AsyncExecutionStatus.SCHEDULED );
        return listener.isScheduled();
      }
    }
  }

tmorgner:
Great find and what a quick fix! Can you create a backlog-case for it, so that we have a JIRA number to go with the fix?

PR #435, opened 2016-09-27
[BACKLOG-11174]-Avoid race condition

EgorZhuk:
@tmorgner please take a look


tmorgner:
Yeah, I could have waited for Wingman, but life is boring if you don't take incredible risks every now and then ...

PR #438, opened 2016-09-29
[BACKLOG-6859]-Row limit in reportviewer

EgorZhuk:
Bringing it back with a default method in listener interface.
With a prayer to CI gods 🙏 
@tmorgner


EgorZhuk:
@tmorgner eventually wingamn's output is the same for the same code

tmorgner:
Let merge, and CI should tell us whether PIR is likely to break on compile. if it does, we should have enough time to patch the beast until the next QAT kicks off.

PR #439, opened 2016-09-29
[BACKLOG-10962] Report Viewer: Images are not displayed in scheduled …

Aliaksandr-Kastenka:
…reports

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/output/PaginationControlWrapper.java
+++ b/src/org/pentaho/reporting/platform/plugin/output/PaginationControlWrapper.java
@@ -32,6 +32,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.URL;
 import java.util.Collections;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


Aliaksandr-Kastenka:
@EgorZhuk @tmorgner 
Please review.

tmorgner:
I dont think you can use "/pentaho" in your RegExp for either CSS or Images. There is no guarantee that the web-application is deployed in that way. 

You'd have to match against "PentahoSystem.getApplicationContext().getFullyQualifiedServerURL()" instead. I would match for "/getImage=" (scrap the "/pentaho" part from the RegExp) and then manually check that the URL matches the fully-qualified URL. Then add in the context path again by removing the protocol, host and port part. 

      URL url = new URL( fullyQualifiedServerUrl );
      String garbage =  url.getProtocol() + "://" + url.getHost() + ":" + url.getPort(); 
      return fullyQualifiedServerUrl.substring(garbage.length());

For our default installations that still should give you "/pentaho/getImage?image=whatever.png". But if installed in a different context, you should get the expected ""/differentContext/getImage?image=whatever.png" as expected.

Aliaksandr-Kastenka:
@tmorgner 
Pull-request updated.

PR #441, opened 2016-10-07
[BACKLOG-10736] - parameter localizations

prakapenka:
... and we need to deeply refactor ParameterXmlContentHandler.class it becomes very big. We need breed parameter validation / dependency discovering / xml generation.
I promise this was last time I had added something without re-writing and decoupling the whole functionality.


prakapenka:
please wingman show some mercy

prakapenka:
@tmorgner  please review

src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
```diff
@@ -738,18 +738,22 @@ Element createParameterElement( final ParameterDefinitionEntry parameter,
...
-            final String attributeValue = parameter.getParameterAttribute( namespace, attributeName, parameterContext );
-            // expecting: label, parameter-render-type, parameter-layout
-            // but others possible as well, so we set them all
-            final Element attributeElement = document.createElement( "attribute" ); // NON-NLS
-            attributeElement.setAttribute( "namespace", namespace ); // NON-NLS
-            attributeElement.setAttribute( "name", attributeName ); // NON-NLS
-            attributeElement.setAttribute( "value", attributeValue ); // NON-NLS
-
-            parameterElement.appendChild( attributeElement );
+            final String attributeValue =
```
tmorgner: The code inside the if-block is the same as in the processTranslateMethod. The only difference is the content of the value-attribute. (I ignore the "TRANSLATE_RESOURCE_ID" attribute behaviour, as there is no harm in including it in the final XML document.)

If you rewrite your "processTranslated" method to return the translated or untranslated text as string, instead of true, you can cut out a few lines of code and the logic flow more readable here.

prakapenka: We need to skip some attributes at all - we don't need translate id on ui, we don't need post processing formula on ui (but it is exposed), we don't need... The boolean value is just to skip attribute with translate id. Is it ok to send to UI unused attributes? That could be rewritten with return string. But ParameterXmlContentHandler already so overload it will not help a lot in general picture.

tmorgner: Yes, it is more than OK to send those attributes. We call it a feature, as it allows OEMs to add their own attributes to parameters to drive their own logic in their parameter UIs.

And yes, that class could need a clean up. The drawback of using DOM is that this code is verbose like hell. But anything we can do to keep the new code as slim, simple and easy to read at least wont make the situation worse.

PR #446, opened 2016-10-12
[BACKLOG-10571] - Previous report job is not cancelled if 'View Repor…

dzmitry-hryhoryeu:
[BACKLOG-10571] - Previous report job is not cancelled if 'View Report' button is pressed.

@tmorgner please review.


dzmitry-hryhoryeu:
@tmorgner please review.

PR #447, opened 2016-10-12
[BACKLOG-11519]-Ignore system row limit if row limit UI is disabled

EgorZhuk:
Ignore any system row limit if the UI is disabled to avoid mess on client side.
@tmorgner please review


tmorgner:
Look sane, and I can't think of anything that Wingman would complain about. Merge.

lucboudreau:
Wingman borked. Started again. Any reasons why this build would hang? Are there any usual suspects we should bring in?

EgorZhuk:
@lucboudreau @tmorgner the test is pretty old and it's the first time it hanged. May be the update of the test libraries causes it and may be it's a broken bit in RAM. I will monitor wingman responses for this project and dig into the problem if it is reproduced.

PR #448, opened 2016-10-12
[BACKLOG-11383] Solving NPE

flbrino:


flbrino:
@mkambol please reaview... Coverage results can't be right

tmorgner:
Dont worry about the coverage. There seems to be a problem with Mockito or any of the other mocking components and/or synchronisation within the coverage tool. If you were to commit an empty line of text, the coverage would still change.

PR #454, opened 2016-10-14
[BACKLOG-11206] - Report Viewer: 'Request failed' message when inputt…

dzmitry-hryhoryeu:
[BACKLOG-11206] - Report Viewer: 'Request failed' message when inputting letters in parameter with numeric value type.

@tmorgner please review.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
+++ b/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
@@ -747,8 +747,7 @@ public class ParameterXmlContentHandler {
 
             parameterElement.appendChild( attributeElement );
 
-            if ( isTextFieldButNotString( valueType, attributeValue ) || ParameterAttributeNames.Core.POST_PROCESSOR_FORMULA.equals( attributeName )
-                    || ( ( ParameterAttributeNames.Core.RE_EVALUATE_ON_FAILED_VALUES.equals( attributeName ) || ParameterAttributeNames.Core.AUTOFILL_SELECTION.equals( attributeName ) ) && "true".equals( attributeValue ) ) ) {
+            if ( isParameterServerValidated( attributeName, attributeValue, valueType ) ) {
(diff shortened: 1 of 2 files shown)


tmorgner:
Expand that test a bit: 
Only evaluate on server when it is any kind of text-field where the parameter-type is not string. This way, ordinary string text fields can still be processed on the client.

dzmitry-hryhoryeu:
@tmorgner please review.

src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
```diff
@@ -747,7 +747,8 @@ Element createParameterElement( final ParameterDefinitionEntry parameter,
 
             parameterElement.appendChild( attributeElement );
 
-            if ( ParameterAttributeNames.Core.POST_PROCESSOR_FORMULA.equals( attributeName )
+            if ( ( ( attributeValue.equals( "textbox" ) || attributeValue.equals( "multi-line" ) ) && !Number.class.isAssignableFrom( valueType ) && valueType != String.class )
```
tmorgner: Dates can be validated on the client too. See my comment on the common-ui pull request for that.

Once a condition spans multiple lines, it is a good time to extract it into a new method: 

protected bool isParameterServerValidated(string attrName, string attrValue) {
  ..
} 

in there you can make the many checks a lot more readable than trying to cram everything into a single statement. 

Also: There are constants defined for the various parameter-presentation types. (ParameterAttributeNames.Core.TYPE_TEXTBOX, etc), please use them so that future maintenance work never has a chance to introduce typos.

tmorgner:
@dzmitry-hryhoryeu A quik fix is needed here and in common-ui to allow date-validation on the client and to break up the large condition here.

tmorgner:
@dzmitry-hryhoryeu I'll merge this as soon as the common-ui part is in, and that one gets merged as soon as it has more tests.

PR #457, opened 2016-10-17
[BACKLOG-11552] Fix hanging executor thread on server shutdown

tkafalas:


tkafalas:
@CodeOnCoffee, please review.  Thanks.

PR #461, opened 2016-10-18
[BACKLOG-11206]-If it's not string - validate it on server

EgorZhuk:
@tmorgner please take a look


tmorgner:
Closing in favour of #462

PR #463, opened 2016-10-21
[BACKLOG-11712]-"View report" button doesn't work after the report ca…

EgorZhuk:
…ncellation

@tmorgner please review


tmorgner:
Looks sane, and Wingman seems to be on holidays, so lets merge.

PR #465, opened 2016-10-25
[BACKLOG-11798] - PIR - After changing the context paths from default…

dzmitry-hryhoryeu:
[BACKLOG-11798] - PIR - After changing the context paths from default, PIR reports no longer function.

@tmorgner please review.


tmorgner:
Good - and just waiting for Wingman.

tmorgner:
@dzmitry-hryhoryeu Please do run the tests before you file the pull request. It should be your default move as soon as you finish coding. Please make the failing test more meaningful by setting up some URL in the ApplicationContext.

PR #466, opened 2016-10-25
[CLEANUP]-Bring the ignored test back to it's duties

EgorZhuk:
Just found that one of the tests was ignored after the Spring version update.
As far as we don't have Spring 2 in the final build, I believe it's preety safe.
@tmorgner


EgorZhuk:
Also we don't need junit here:)

tmorgner:
Looks sane, but lets wait for Wingman so that we dont become roadkill because of some strange stuff going on. Days before the release I wont take risks :)

EgorZhuk:
@tmorgner I'm not going to cherry-pick it to the release or sp branch. But it may hit somebody in future, so it's worth to have it in master.

PR #470, opened 2016-10-28
[PRD-3883]-Support DRILLDOWN links from ChartExpressions

EgorZhuk:
Eventually found time for it.

@tmorgner please review


EgorZhuk:
Hm, will check it

tmorgner:
@EgorZhuk Linux vs Windows: Your file name case does not match the actual file name stored in Git. Wingman is Linux, so case does matter.

EgorZhuk:
Thank you. Didn't notice that names are different. Updated the commit.

tmorgner:
Sweet. Merged.

PR #480, opened 2016-11-01
[BACKLOG-9586] - add runAsUser to be able to run with discarded session

prakapenka:
The trick is when user do logout server explicitly call discard on a session object. If under the IPentahoSession is HttpSession exactly http session is discarded. Move over, after a session is discarded report processor attempts to load mondrian cube from repository. Using AOP spring intercept such call and try to validate it according to a user session - but it is discarded since user had logout. So - we cant use http session to authorize such calls, instead, we should use user name instead of obsolete session object. See BACKLOG-9586 for detailed stack-trace (there is really self-exploratory stack-trace).

### DIFF
--- a/test-src/org/pentaho/reporting/platform/plugin/connection/PentahoMondrianConnectionProviderTest.java
+++ b/test-src/org/pentaho/reporting/platform/plugin/connection/PentahoMondrianConnectionProviderTest.java
@@ -12,26 +12,50 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
  */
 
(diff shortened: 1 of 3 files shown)


tmorgner:
@prakapenka Something blew up. Mondrian user roles trigger an NPE. Seems related to your fix.

prakapenka:
We don't have sessions in unit tests, we don't have security context in unit tests, so why not finally get an NPE if you did not to manage to mock all this stuff there? My fault, did not run unit tests, just check it directly on server.

test-src/org/pentaho/reporting/platform/plugin/connection/PentahoMondrianConnectionProviderTest.java
```diff
@@ -12,26 +12,50 @@
...
+  @AfterClass
+  public static void afterClass() {
+    PentahoSessionHolder.setSession( null );
+  }
 
-  protected void setUp() {
-    provider = new PentahoMondrianConnectionProvider();
+  @Before
+  public void before() {
+    when( session.getName() ).thenReturn( "karasik" );
```
prakapenka: Karasik should satisfy NPE in SecurityHelper test.

PR #487, opened 2016-11-03
[BACKLOG-11427] Report Viewer: Row Limit control is not disabled if R…

Aliaksandr-Kastenka:
…eport Row Limit was set using OpenFormula


Aliaksandr-Kastenka:
@tmorgner 
Please review.

PR #488, opened 2016-11-03
[BACKLOG-12100]-Clear cache for the current user only

EgorZhuk:
@tmorgner could you please review

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
+++ b/src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
@@ -38,11 +38,10 @@ public class CacheManagerEndpoint {
     try {
       final IPluginCacheManager iPluginCacheManager = PentahoSystem.get( IPluginCacheManager.class );
       final IReportContentCache cache = iPluginCacheManager.getCache();
-      cache.cleanup();
+      cache.cleanupCurrentSession();
 
       final ICacheManager cacheManager = PentahoSystem.get( ICacheManager.class );
 
-      cacheManager.clearRegionCache( "report-dataset-cache" );
       cacheManager.clearRegionCache( "report-output-handlers" );
 
       DataCacheFactory.getCache().getCacheManager().clearAll();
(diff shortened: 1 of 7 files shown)


dkincade:
@CodeOnCoffee please review this update to cache management for your thoughts ... @kcruzada has details

src/org/pentaho/reporting/platform/plugin/cache/DeleteOldOnAccessCache.java
```diff
@@ -120,6 +120,10 @@ protected void setMillisToLive( final long millisToLive ) {
     cleanUp();
   }
 
+  @Override public void cleanupCurrentSession() {
+    cleanup();
```
tmorgner: This performs a global cleanup. It does not just limit itself to the current session.
If you want to limit to the current session the list in the "purgeSegment" call below needs to have two entries: SEGMENT and session.getName().

EgorZhuk:
Will update the PR soon.

src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
```diff
@@ -38,15 +37,13 @@ public Response clear() {
     try {
       final IPluginCacheManager iPluginCacheManager = PentahoSystem.get( IPluginCacheManager.class );
       final IReportContentCache cache = iPluginCacheManager.getCache();
-      cache.cleanup();
+      cache.cleanupCurrentSession();
 
       final ICacheManager cacheManager = PentahoSystem.get( ICacheManager.class );
 
       cacheManager.clearRegionCache( "report-dataset-cache" );
```
tmorgner: This is a global purge across all sessions.

src/org/pentaho/reporting/platform/plugin/CacheManagerEndpoint.java
```diff
@@ -38,15 +37,13 @@ public Response clear() {
...
       cacheManager.clearRegionCache( "report-output-handlers" );
 
-      DataCacheFactory.getCache().getCacheManager().clearAll();
```
tmorgner: This could be a local purge, assuming the PentahoDataCache implementation is changed to not wipe the region. 

see PentahoDataCache.PentahoDataCacheManager#clearAll and PentahoDataCacheManager#shutdown
The clearAll call there should call "killSessionCache" instead.

tmorgner:
Looks good now. Safe and sane.

@CodeOnCoffee Any objections?

tmorgner:
Wingman is happy too.

pentaho-nbaker:
Looks good to me

PR #491, opened 2016-11-03
[BACKLOG-12100]-Clear cache for the current user only

EgorZhuk:
#488 port @tmorgner

reopening for Wingman


tmorgner:
@wingman-pentaho This bot is playing Terminator and does not let this blocker-bug fix pass. Can we sent someone back in time to prevent Wingman from becoming evil? Or lacking a time machine, can you clean the ivy cache so  that wingman can build happily?

lucboudreau:
You can clear anything you need by using the job 'wingman-clear-artifact' on Wingman.

PR #495, opened 2016-11-09
[BACKLOG-12234]-Fix fast csv output + tests

EgorZhuk:
Epic bug that should be caught by tests.
Fixed it and updated tests to touch fast mode logic.

@tmorgner


tmorgner:
Seriously? I'd shoot the author of that class, but then suicide is against catholic rules, so I'll leave that idea! :)

Nice find.

EgorZhuk:
@tmorgner then shoot me because I'm the bugger 🔫

tmorgner:
wow, that sort of bug is normally my speciality. If you now also start to mess up seemingly simple boolean checks, then consider yourself infected. There will be no cure :)

PR #499, opened 2016-11-09
[BACKLOG-12234]-Fix fast csv output + tests

EgorZhuk:
Hope wingman would run again.

@tmorgner


EgorZhuk:
@tmorgner it's ok now

PR #500, opened 2016-11-23
PPP-3592 Report parameter vulnerable to XSS

denisprotopopov:
- added encoding for html
 - added unit test


denisprotopopov:
@pamval @tmorgner please review the changes

tmorgner:
The parameter XML is a general OEM interface. You cannot/shouldnot simply assume that it is only used for the report-viewer. 

HTML/JavaScript has means of securely displaying messages, so the real fix would be to actually fix the PromptPanel/Error component in common-ui to process the message in a secure way. There either avoid setting the text as "innerHTML" (a bad practice so evil that it should be banned!), or encode the text before setting it with that silly method. 

Given that CDE defaults to the insecure ".html" in its TextComponent, escape the text as seen in LabelBuilder.js in common-ui.

PR #501, opened 2016-11-24
[BACKLOG-11526] - An arrow is absent on Row Limit drop-down.

dzmitry-hryhoryeu:
[BACKLOG-11526] - An arrow is absent on Row Limit drop-down.


dzmitry-hryhoryeu:
@tmorgner please review.

PR #509, opened 2016-12-16
BISERVER-13435 Cascading parameter retains old values and shows key-value from old values

AliaksandrDrebenchuk:
selection process was fixed


kolinus:
@tmorgner @pamval could you please take a look

tmorgner:
Looks good. Now all I need here is an actual unit-test that validates that the fix works. This should be an easy task, as the class in question produces regular XML that can be parsed and validated easily.

AliaksandrDrebenchuk:
@tmorgner unit-test was added

PR #510, opened 2016-12-21
BISERVER-13435 Cascading parameter retains old values and shows key-value from old values

AliaksandrDrebenchuk:
getSelections method was fixed


AliaksandrDrebenchuk:
@tmorgner @pamval could you please take a look this small addition for https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/509

PR #511, opened 2017-01-04
[BISERVER-13435]-fix regression

EgorZhuk:
While validating PDB-1931 I found a regression caused by BISERVER-13435 fix.
Basically the fix overrides all parameters values disregarding of is it actually a list parameter or not.
With such logic we will lose the default values for the plain parameters.

@AliaksandrDrebenchuk FYI
@tmorgner please review


tmorgner:
Nice catch that!

PR #512, opened 2017-01-09
BISERVER-13170-Always send post-processing parameters values to the server

EgorZhuk:
The 1st commit (java code) is a fix for the stackowerflow which is a regression after Ewok's work.
The 2nd is actually the bug fix. We don't reload the PromptPanel after 6.1 and even don't send a request to server after 7.0, so to show valid value we need to go and set it through widget API.
I'm resetting datespickers only here but probably we want the same logic for other inputs as well.
Please pay attention to date parsing, I assume we always have date in ISO format there.

### DIFF
--- a/package-res/reportviewer/reportviewer-prompt.js
+++ b/package-res/reportviewer/reportviewer-prompt.js
@@ -299,10 +299,7 @@ define([
           if(oldParams){
             oldParams.mapParameters( function (p) {
               if(names && names.indexOf(p.name) >= 0){
-                //Don't need to request formulas
-                if(p && (!p.attributes || (p.attributes && !p.attributes['post-processor-formula']))){
                   needToUpdate.push(p.name);
-                }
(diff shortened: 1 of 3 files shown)


EgorZhuk:
Depends on https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/919

EgorZhuk:
[NOT FOUND  ] pentaho-kettle#kettle-engine;7.1-SNAPSHOT!kettle-engine.jar(source) (0ms)
The kettle is boiling wingman's mind

EgorZhuk:
@tmorgner also green now

EgorZhuk:
Checked locally, tests pass, amended the commit

EgorZhuk:
He was bored and just waited for a joke before going back to his duties.

tmorgner:
I hate him too :)

PR #513, opened 2017-01-13
[BISERVER-13407]-ValidationQuery and Pooling options not sent to DB when datasource is defined in Manage DataSources Perspective

denisprotopopov:
- added receiving connection from PoolingDataSource
 - added unit tests

PR #514, opened 2017-01-16
[SP-3101][BISERVER-13407]-ValidationQuery and Pooling options not sen…

denisprotopopov:
…t to DB when datasource is defined in Manage DataSources Perspective

 - added receiving connection from PoolingDataSource
 - added unit tests

It should be merged atfer this https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/513


tmorgner:
Merged. And given that this is the same code, I think it is safe to tell Wingman to stuff it and merge the master one as well.

PR #515, opened 2017-01-16
[SP-3102][BISERVER-13407]-ValidationQuery and Pooling options not sen…

denisprotopopov:
…t to DB when datasource is defined in Manage DataSources Perspective

 - added receiving connection from PoolingDataSource
 - added unit tests

It should be merged after this
https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/513


tmorgner:
@denisprotopopov There are compile errors here. Probably because of older mocking libraries or something equally silly. Please adjust the test-case code to make it work.

PR #520, opened 2017-01-25
[SP-3236][BISERVER-13170]-Avoid Stackoverflow on cyclic parameter dep…

EgorZhuk:
…endencies.

Always send post-processing parameter selections to the server.


EgorZhuk:
@mchen-len-son @pamval @wingman-pentaho same CCC isue

mchen-len-son:
Local build is OK. All the tests are passing. No checkstyle violations.

PR #521, opened 2017-01-25
[SP-3178][BISERVER-13435]-Cascading parameter retains old values and …

EgorZhuk:
…shows key-value from old values

The simplier version as far as partially sent parameter xmls are not implemented in 6.1


EgorZhuk:
@tmorgner 6.1 is broken on Jenkins since Jan 17. @denisprotopopov seems you need to update some tests for https://github.com/pentaho/pentaho-platform-plugin-reporting/commit/fd7929d9612d1a977fcfc2483c1f7f8088e54f23 see the job results here  http://10.177.176.19/job/6.1-snapshot/job/platform-reporting-plugin/

PR #525, opened 2017-01-27
[BISERVER-12804] PDF-Output is not shown in PUC 5.2 in Internet Explorer

tim-ryzhov:
* The issue is reproduced only in IE11 and PDF Output
* onLoad event is not triggered when hidden form is submitted to iframe
* Using  "onreadystatechange" <iframe> attribute to trigger the event

### DIFF
--- a/package-res/reportviewer/reportviewer.js
+++ b/package-res/reportviewer/reportviewer.js
@@ -1116,7 +1116,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
                     me._isFinished = true;
                     me._hideAsyncScreens();
                     //Let's get first page
-                    me._getContent(firstPageJobStatus.uuid, url, function () {
+                    me._getContent(firstPageJobStatus.uuid, url, null, function () {
                       isIframeContentSet = true;
                     });
 
@@ -1189,7 +1189,7 @@ define([ 'common-ui/util/util', 'common-ui/util/timeutil', 'common-ui/util/forma
 
                   //Show report
                   if (!isIframeContentSet) {
-                    me._getContent(mainJobStatus.uuid, url, function () {
+                    me._getContent(mainJobStatus.uuid, url, mainJobStatus.mimeType, function () {
                       isIframeContentSet = true;
                     });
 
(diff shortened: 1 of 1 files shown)


EgorZhuk:
@tim-ryzhov looks that your fix rely on an async mode which was introduced in a 7.0 and is actually a switchable feature. Could you please set async = false in a plugin.settings.xml and check if your fix stil works?

tmorgner:
Egor is right. You only fix half the issue. And the fact that it is reported against 5.x tells me that your fix does not apply to the problem there. 

On async-mode set to false, we still use the hidden form to submit the report, but you will receive no events about the state of the report processing. The hidden frame will receive PDF content, which then is handed off to a IE PDF plugin (not sure whether they still use ActiveX, but that used to be a source of troubles. AcrobatReader in generally is a steaming pile of [censored] in that respect). 

Our iframe has no event handler for the "load" event if you are in IE - we use the "readystate" event handler instead. According to MSDN, in IE-11 you are supposed to use the load event instead. 
https://msdn.microsoft.com/en-us/library/ms534359(v=vs.85).aspx

So by not using "readystate" and using "load" when on IE11 you may be able to fix the problem without having to resort on that hackish IE11 workaround you use for the async mode.

Given that your case has first been flagged up in 5.x, and given that the report-viewer has changed dramatically since then, you will find that you cannot apply the exact same fix on both places.

Additionally, the first comment on the case seems to point to an encoding problem. In 5.x we did not use a hidden frame. The iframe there was triggered by simply setting the "src" URL with all parameters attached to that URL. According to Jim's comment, the encoding used for producing that URL is wrong.

tim-ryzhov:
That's right, the fix doesn't work in synch mode. It should be updated.
In the current implementation IE11 binds the "load" event,  as dojo   has("ie") method returns "underfined" for IE11 [http://dojo-toolkit.33424.n3.nabble.com/Dojo-1-9-2-now-available-td4000355.html](http://dojo-toolkit.33424.n3.nabble.com/Dojo-1-9-2-now-available-td4000355.html)
Next point is that the "load" event is not fired in  iframe in IE11. This problem is mentioned here: [https://connect.microsoft.com/IE/feedback/details/809377/ie-11-load-event-doesnt-fired-for-pdf-in-iframe](https://connect.microsoft.com/IE/feedback/details/809377/ie-11-load-event-doesnt-fired-for-pdf-in-iframe)
Also, in 7.1 encoding issue is not reproduced
The same hackish approach can be done for sync mode, as I can see it now.

tim-ryzhov:
Found another approach.  "Onreadystatechange" <iframe> attribute is used to trigger the event with jquery

tmorgner:
Wow, that code feels dirty, yet at the same time it is a sane approach. It says a lot about why I love IE. I bet full mastery of IE quirks means eternal job-security and a warm place in an asylum.

PR #526, opened 2017-01-27
[SP-3178][BISERVER-13435]-Backport dependent parameters logic to fix …

EgorZhuk:
…cascading parameter retains old values.

The original backport didn't work. To implement a proper solution we need to also backport some stuff from the 7.0. We need to send changed parameters from the ui and apply the logic only to the dependent parameters of those which were sent. Otherwise we will reset parent selections on child changes.


EgorZhuk:
@tmorgner Nice, nonexistent tests failed

tmorgner:
Ignoring bad wingman once again.

PR #527, opened 2017-01-30
[SP-3237][BACKLOG-14238]-Move synchronisation of a parameter definiti…

EgorZhuk:
…on and a dashboard to the reporting plugin

#523 backport

In 6.1 we need to store parameters which we are forcing to update, as long as we don't store the whole parameter definition. Also operation and event API is not implemented there yet.. Other than that it's the same as for 7.0 and 7.1


EgorZhuk:
There are no js tests in 6.1. Let's address it separately.

PR #529, opened 2017-02-06
[BISERVER-13435]-Reset invalid selections for child parameters presen…

EgorZhuk:
…t in a change parameters list if their parent is also present

The case is reproducible when we recieve the parameter, we are going to reset in the changed parameters list . This happens after you change a child parameter. This cahnge doesn't trigger a server call. After that, when changing a parent we have it's child on the list (cause yes, it has changed since the last server sync). It doesn't break any logic on the server (cause the child will be picked up anyway), besides the invalid selection erasure.

This could be fixed on the UI but I prefer to adjust the logic on the server because it is more 1) more secure 2) can be better covered by unit tests.

I've added 3 tests: one for the  parent only present in changed parameters, another for both child and parent and the last for the child and parent but on the deeper dependency level to ensure that we don't erase all children but only those which have a parent in the changed parameters list.

### DIFF
--- a/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
+++ b/src/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
@@ -501,8 +501,11 @@ public class ParameterXmlContentHandler {
 
                   .toList() );
 
+      //We only need the top level parents in order to erase children selections in lists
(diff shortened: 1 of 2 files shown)


pamval:
@tmorgner can you review this one please?

tmorgner:
Egors reasoning seems sound - both for the code fix approach and for wanting to do this on the server. Good to go.

PR #532, opened 2017-02-08
[BACKLOG-14580]-Don't reset selection on the report opening

EgorZhuk:
One case has slipped out of our sight (and even was covered by a unit test). On the initial parameter call we don't have anything changed but already try to erase something. It causes the erasure of the top level parameter if: a) it is a list b) it has no validation.


pamval:
@tmorgner can you get this one, please?

PR #536, opened 2017-02-15
[PRD-5101]-Show a page control only for a paginated HTML output type

EgorZhuk:


EgorZhuk:
Ok, let's add another mock

PR #537, opened 2017-02-16
[SP-3237][BISERVER-13170]-Reset list selections only when force updat…

EgorZhuk:
…e is required

On 6.1 setting the flag isn't triggering an update for a list parameter. We need a difference in selected values.
Turns out, we can't simply select values in _applyUserInput, because it will also hide a selections for a parameter without PP formula. 
Let's revert the logic for lists without PP formulas. 

To trigger an update, let's reset values for lists with a post processing which we do want to update forcefully.


EgorZhuk:
There is no JS tests in 6.1

PR #539, opened 2017-02-17
[SP-3262][BISERVER-12804] PDF-Output is not shown in PUC 5.2 in Inter…

tim-ryzhov:
…net Explorer 11

* In case of IE11 and PDF report as onLoad event is not triggered
* Added workaround by  triggering event once, using "onreadystatechange" <iframe> attribute


tim-ryzhov:
This is a backport of #525 and #528

tmorgner:
Yes,the same code, so merge is good.

PR #540, opened 2017-02-17
[CLEANUP]-Empty JS test suite to avoid Wingman errors

EgorZhuk:
Let's make Wingman happy.


tmorgner:
So the nagging spouse theory is true - at least for Wingman. Nag long enough and someone will clean out the garbage. :)

PR #543, opened 2017-02-22
[PPP-3410]-Additional cleanup on the UI

EgorZhuk:


EgorZhuk:
No tea if there is no kettle!

PR #546, opened 2017-03-09
[SP-3349][BISERVER-13544]-Protect parent/top calls with an isRunningI…

EgorZhuk:
…FrameInSameOrigin check


EgorZhuk:
The PR doesn't contain Java code changes and compiles locally, so it's a wingman fluctuation.

mchen-len-son:
I think, wingman works correctly. The failure is probably connected with the latest mavenization updates

PR #548, opened 2017-03-17
[SP-3342][PRD-5101]-Show a page control only for a paginated HTML out…

EgorZhuk:
…put type

#536 backport


EgorZhuk:
Works locally. Probably it's time to invest in a CCC issue resolution.

PR #549, opened 2017-03-20
[PIR-1326] add date for parameter xml

alina-ipatina:
@mchen-len-son @EgorZhuk @tmorgner please review.

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
@@ -896,6 +896,7 @@ public class ParameterXmlContentHandler {
         final Element valuesElement = document.createElement( "values" ); //$NON-NLS-1$
         parameterElement.appendChild( valuesElement );
 
+        boolean valuesContainDateType = false;
         final ParameterValues possibleValues = asListParam.getValues( parameterContext );
         for ( int i = 0; i < possibleValues.getRowCount(); i++ ) {
           Object key = possibleValues.getKeyValue( i );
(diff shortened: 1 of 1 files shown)


tmorgner:
I dont think this is correct. 

I can see even without looking at the non-existing test code, that your code does not fix the problem. You merely rewrite what the current code does already. Then you blindly, and rather in a rude fashion attempt to hardcode date-type into the parameter if you encounter a date anywhere in the selected values.

But as far as I can tell, the problem here is not about selections, it is about how the UI displays dates.

Looking at the case and what the expected behaviour is, I would say you are "fixing" the wrong spot.

You should be fixing the PIR-UI, or the way the parameter is defined when a date-field is dragged into the filter section. 

If the parameter has a format-mask defined, the format mask should have been used to display the values. PIR is notorious with not matching model values with the state of the GUI - so that would be the first thing to check. When I look at the screenshots given in BACKLOG-5975, it looks as if the parameter should have had a "yy-MM-dd" format applied as default. 

If the parameter has no format mask defined, then things get a bit tricky:

If the field-type is **java.util.Date**, then you have to assume that the date object is actually a time-stamp. The older Java API for dates is badly defined, and Date does not mean the same as the word "Date" in English. 

Dates always carry a time-stamp, as the date is just a wrapper around a C-style Date (milliseconds since 1.1.1970). Other people have written [long essays](http://stackoverflow.com/questions/1969442/whats-wrong-with-java-date-time-api) about what is wrong with that sort of definition, so I wont repeat that here. 

If the field is a **java.sql.Date**, then yes, this is a days-only date and the timestamp part can be ignored.

If the field is a **java.sql.TimeStamp** then you are back at treating it as a datetime object, same as java.util.Date.

If the parameter has no formatting mask, then depending on the data-type used, you could default the parameter to a suitable formatting mask.

In no case would you change how dates are transfered over the network or how the server side interprets these dates. We transfer all dates (Timestamp and sql-Date) as ISO-8601 formatted date. This is a safe, well defined format and there is no need to change anything on that. Formatting is independent of the data transfer of values.

alina-ipatina:
@tmorgner thank you for review, could you please answer some questions.

During parameter creation parameter's type obtained from metadata as "DATE" and mapped to java.sql.Timestamp:
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/master/web/src/main/javascript/pir/pir/controller.js#L3516
https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/blob/master/web/src/main/javascript/pir/pir/controller.js#L86
As far as I can see there is no information in PIR UI about actual date type - is it java.sql.Timestamp or java.sql.Date.

What field do you mean in your notes? Date fields in model are java.sql.Timestamp by default.

PR #552, opened 2017-03-24
[BISERVER-13569]-Compare values in datepicker parameters carefully

EgorZhuk:
Enhanced the logic and covered it with tests.
There were several issues:
1) The timezone was added twice, when it was present in the value and the hint was aldo present.
2) Some Ausie and New Zeland timezones blowed things up. For some reason the format for 30/45 minutes timezones is different and not supported by JS dates. We have +09.530 and +11.7545 against regular +0930 and +1145 for Australian Nothern territory and NZ-CHAT timezones. By the way, the values with those timezones are not displayed in PUC ssatrting from a 6.1 GA. I've opend an issue PRD-5879
3) If the date is invalid we won't loop infinitely

### DIFF
--- a/core/src/main/javascript/reportviewer/reportviewer-prompt.js
+++ b/core/src/main/javascript/reportviewer/reportviewer-prompt.js
@@ -651,19 +651,62 @@ define([
           }
         } else {
           //Working with a plain parameter
-          //Already has a valid value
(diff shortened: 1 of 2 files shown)


pamval:
@diogofscmariano @afrjorge @annapopova1 please review

PR #563, opened 2017-04-25
[PRD-5879] Datepicker doesn't display a user selection if a date has a 30/45 minute time zone

Advent51:
- Offset formatting was fixed


Advent51:
@mbatchelor @tmorgner please review

Advent51:
This test successfully passed locally

tmorgner:
Dont worry about the test - there must be some flaky race condition or deadlock going on. We had similar troubles with that part of the test suite for quite a while.

You are only changing JS files, so if a Java test breaks as a result as your commit, you can be sure that your change is not the cause of it.

The file you changed is the exact same copy as the one in common-ui, so as long as the common-ui one has the tests, that should be fine. I created [BACKLOG-16159](http://jira.pentaho.com/browse/BACKLOG-16159) to simply remove the duplicate code. As my granny says: No code, no bugs.

PR #573, opened 2017-06-01
[BISERVER-12968] - Self Links defined in a PRPT chart do not work on the

PashkovichAndrei:
server

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/PentahoReportEnvironment.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/PentahoReportEnvironment.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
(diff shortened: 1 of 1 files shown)


PashkovichAndrei:

tmorgner:
Make some changes to the "PentahoReportEnvironment" class here so that this class can correctly return the caller URL. 

How? Simple: First extend that class so that it accepts an repository-path as parameter in a new constructor. Mark the old constructor as deprecated.  Change the class so that it returns that path on the URL when asked for "selfURL". If no repository-path is set, return null. Be aware that it is possible for reports to not have a valid path. The URL you should return looks like ``fullyQualifiedServerUrl + "api/repos/" + encodedReportPath + "/viewer"``. 

Your current return value on "selfURL" will not produce a valid call for the report-viewer. I bet you did not test this, it obviously misses the path element.

Next, change both SimpleReportingAction and SimpleReportingComponent so that they supply a correct repository-path for the processed report to the new constructor. Construct the path from the information you have about the report at that time. If the report has been loaded from the JCR, use that path. When you look at SimpleReportingAction#getReport you may notice that there are cases where the report is loaded from a stream. For this case you don't have a valid URL, and therefore pass null as URL into the constructor.

That should now be able to see a valid self-url links for all reports that do self-linking.

And once that is done, and there are test-cases for this case as well (never ship new code without covering with tests!) make sure you also cover the interactive-reporting plugin in a similar fashion.

PR #574, opened 2017-06-14
[BISERVER-13663] Report goes into loop when date picker uses server t…

mchen-len-son:
…imezone after applying latest service packs

 - added date-only comparison for datepicker values to findChangedParameters func


mchen-len-son:
@tmorgner can you please take a look?

PR #575, opened 2017-06-16
[SP-3647][BISERVER-13663] Report goes into loop when date picker uses…

mchen-len-son:
… server timezone after applying latest service packs

 - added date-only comparison for datepicker values to findChangedParameters func


mchen-len-son:
@pamval @kolinus please take a look
thanks.

PR #577, opened 2017-06-16
[BISERVER-12968]-Self Links defined in a PRPT chart do not work on the

PashkovichAndrei:
server

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.reporting.platform.plugin;
@@ -46,6 +46,7 @@ import org.pentaho.reporting.engine.classic.extensions.modules.java14print.Java1
 import org.pentaho.reporting.libraries.base.util.CSVQuoter;
 import org.pentaho.reporting.libraries.base.util.StringUtils;
 import org.pentaho.reporting.libraries.resourceloader.ResourceException;
+import org.pentaho.reporting.libraries.resourceloader.ResourceKey;
 import org.pentaho.reporting.libraries.xmlns.common.ParserUtil;
 import org.pentaho.reporting.platform.plugin.cache.NullReportCache;
 import org.pentaho.reporting.platform.plugin.cache.ReportCache;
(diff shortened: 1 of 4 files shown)


PashkovichAndrei:

pamval:
@tmorgner, please review when you get the chance

PashkovichAndrei:
@tmorgner, @pamval, @mbatchelor, could you please do review?

core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -588,6 +591,14 @@ private String extractContentLinkSpec() {
     }
   }
 
+  static String getReportPath( MasterReport report ) {
+    ResourceManager rm = report != null ? report.getResourceManager() : null;
+    BundleResourceManagerBackend backend = rm != null ? (BundleResourceManagerBackend) rm.getBackend() : null;
```
tmorgner: You cannot assume that in 100% of the cases. Some weird person may actually call this action/component with a self-constructed MasterReport instance that has not been loaded from a bundle. Use an "instanceof" check here.

You may find it safer to work directly with the resource-key itself. Have a look at the while-loop in "PentahoKettleTransFormFileProducer#computeFullFilename". Combined with a similar loop in AbstractKettleTransformationProducer#computeFullFilename" testing for java.io.File objects, this is the safest solution to the problem and handles all sort of weird side-cases (like loading a prpt-report contained in a zip or jar file).

PashkovichAndrei:

tmorgner:
Let me reapeat this outside of a hidden "code review" comment: 

Adding the "instanceof" check is a good example for a "bare minimum fix". Please adapt your code to locate the resource-key in the same way the KettleTransformationProducer does it. Copying that while-loop is not that hard!

For reference: This is the earlier comment:
You may find it safer to work directly with the resource-key itself. Have a look at the while-loop in "PentahoKettleTransFormFileProducer#computeFullFilename". Combined with a similar loop in AbstractKettleTransformationProducer#computeFullFilename" testing for java.io.File objects, this is the safest solution to the problem and handles all sort of weird side-cases (like loading a prpt-report contained in a zip or jar file).

PashkovichAndrei:

core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -588,6 +592,24 @@ private String extractContentLinkSpec() {
     }
   }
 
+  static String getReportPath( MasterReport report ) {
```
tmorgner: Try "report.getResourceKey()" for a change. That will give you the resource-key the report was loaded with without having to do all that unwrapping. Remember: If it is awfully complex, then you might be doing it wrong. 

    ResourceKey key = report.getResourceKey();

can replace all your first three lines of code, and you dont have to blind-cast the key-variable into a ResourceKey any more. How cool is that?

PashkovichAndrei:

PashkovichAndrei:
@tmorgner could you please have a look?

tmorgner:
That looks good now.

PR #583, opened 2017-09-19
[BACKLOG-18902] Reporting Prompts Button controls are different size height between Ruby and Sapphire

davidmsantos90:
@pentaho/millenniumfalcon please review


davidmsantos90:
@pentaho/millenniumfalcon locally every test runs ok, and also I didn't change any code, so this wingman error doesn't make sence

PR #588, opened 2017-09-26
[CLEANUP]: Ignore settings for Eclipse IDE.

TatsianaKasiankova:


tmorgner:
No functional change, and thus ignoring wingman.

PR #589, opened 2017-09-28
[BISERVER-13685]: The certain report with Post Processing Formulas and data pickers goes into a loop and it never stops

TatsianaKasiankova:
The problem arises on changing value in data-picker parameter at the certain conditions:
1)Parameter has data-format;
2)Parameter has post-processor-formula;
3)Post-processor-formula generates the date in the format that differs from data-format;

Parameter xml generated for Report Viewer contains date in the internal transport format (ISO-8601) used by Pentaho Reporting Engine.
But also parameter can contain its own data-format.
We need to convert dates from/to these formats to compare them correctly.
Now we do not do this one.
That is why on comparying "old" and "new" data in the function _compareDatesOnly we compare two different dates in fact .
And of course they are not equall and we try to set up the "new" value again and again.

The fix is to apply parameter's defined format and the transport formatters to "old" and "new" dates before their comparying.

### DIFF
--- a/core/src/main/javascript/reportviewer/reportviewer-prompt.js
+++ b/core/src/main/javascript/reportviewer/reportviewer-prompt.js
@@ -23,8 +23,9 @@ define([
   "common-ui/jquery-clean",
   "common-ui/underscore",
   "cdf/dashboard/Utils",
-  "dojo/dom-class"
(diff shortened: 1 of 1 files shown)


TatsianaKasiankova:
@tmorgner , @pamval Hi, please review the PR.

pamval:
@diogofscmariano @afrjorge @annapopova1 please review

PR #594, opened 2017-10-10
[BISERVER-13706]-PRPT Report Run From Pentaho Server Fails in White S…

PashkovichAndrei:
…creen

After Initial View Report Run Using Several Output Types


PashkovichAndrei:

tmorgner:
Looks good. That part of the dispose seems overly eager, and would misbehave when the contentframe is not used for actual content (as in the excel export case). Any references should be cleaned up when the reporting tab itself is closed anyway.

PR #595, opened 2017-10-11
[PRD-5915] 404 opensans-regular font not found error when running prpt report in 7.1

Advent51:
- css file was excluded due to the load issues


Advent51:
@pamval @dcleao @mchen-len-son please review

PR #599, opened 2017-10-30
[SP-3822]-[BISERVER-13706]-PRPT Report Run From Pentaho Server Fails …

PashkovichAndrei:
…in White Screen

After Initial View Report Run Using Several Output Types


mchen-len-son:
Looks like those are indeed non-deterministic tests - could not reproduce locally

PR #603, opened 2017-11-14
[BISERVER-13744]-Security popup appear when changing a param on a report

stepanovdg:


stepanovdg:
@tmorgner @dmariano Could you please take a look?

mchen-len-son:
@pamval @tmorgner @mbatchelor can you please take a look?

PR #607, opened 2017-12-21
[BISERVER-13735] Run in Background Dialog should not be shown when PRPT and PRPTI reports are being used on a Dashboard

Advent51:
- restriction became stronger


Advent51:
@afrjorge @mdamour1976 @pamval please review

Advent51:
@afrjorge @mdamour1976 @pamval please review

PR #608, opened 2017-12-26
[SP-3973][BISERVER-13744]-Security popup appear when changing a param on a report

stepanovdg:


stepanovdg:
@mchen-len-son Could you please take a look?

PR #610, opened 2018-01-17
[BACKLOG-17248] Fix images/CSS in single-page HTML reports generated in background

pentaho-whartman:
JIRA: http://jira.pentaho.com/browse/BACKLOG-17248

Changes the way repository paths are handled for `ReportContentItems` to work around Spring not allowing encoded slashes.


pentaho-whartman:
Fixed checkstyle issues

PR #614, opened 2018-02-19
[BISERVER-13823]-The prpt report goes into a loop and it never stops …

SergeyTravin:
…when

server and web client are in different time zones and when the report has
a few post-processing formulas


SergeyTravin:
@mchen-len-son review please

SergeyTravin:
@mchen-len-son, @mbatchelor, @pamval please review it

SergeyTravin:
@pentaho-lmartins , please take a look

PR #615, opened 2018-03-01
[BACKLOG-20957] Store images/css in the repository for pageable HTML output when the useJcr flag is set

pentaho-whartman:
@pentaho/rogueone @tmorgner 

JIRA: https://jira.pentaho.com/browse/BACKLOG-20957

This PR modifies `CachingPageableHtmlOutput` to respect the useJcr flag. The `SimpleReportingAction` class was changed to enable this flag by default (so prpts that are run in the background will store images/css in JCR).

Storing related files in a folder with the prpt name was considered, but not implemented. Since images are reused if the report is cached, output may refer to images from an earlier run the prpt. If the cache is cleared, the images will not be deleted and duplicates will be added to the same folder.  This leads to same confusion as having no folder -- as long as images are reused between runs, the dependencies will be complex. This is really a stopgap solution. A long term solution is captured in [this ticket](https://jira.pentaho.com/browse/BACKLOG-21877).

If testing, the reporting cache will need to be cleared (Tools -> Refresh -> Reporting Data Cache). If it is not cleared, any report that have been run with pageable html output will fail.

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingAction.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingAction.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+ * Copyright (c) 2002-2018 Hitachi Vantara.  All rights reserved.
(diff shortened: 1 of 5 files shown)


pentaho-whartman:
Fixed the checkstyle issues. Tests look fine locally and aren't newly broken -- I think that's safe to ignore.

pedrofvteixeira:
PR looks good by me. However, I'm not seeing UTs / ITs accompanying these changes to
- a) prove out the intended behaviour
- b) prevent this behaviour from ever regressing, should future code changes be made by someone else

PR #616, opened 2018-03-16

CravoBranco:

* Fixed report viewer when client and server are in different timezones and DatePicker is configured as "Use Client timezone".


CravoBranco:
@pentaho-lmartins , @mchen-len-son , @mbatchelor , @tmorgner , @duarteteixeira  please review and merge if it's all ok.

tmorgner:
This most definitely needs some unit tests demonstrating the correct parsing and thus also documenting why this code is correct. Otherwise the next time there is an problem we can't tell whether that new change broke something.

CravoBranco:
This solution wasn't good and a new one is being implemented, this time on client side.
So I will close this PR and create a new one.
Thanks for your comments!

PR #618, opened 2018-03-22
[SP-4218][BISERVER-13823]-The prpt report goes into a loop and it nev…

SergeyTravin:
…er stops when

server and web client are in different time zones and when the report has
a few post-processing formulas


SergeyTravin:
@pentaho-lmartins, @mbatchelor, @mchen-len-son please review It is backport for https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/614

PR #620, opened 2018-03-26
[BISERVER-13895]-Cannot read property 'fireOutputParam' of undefined when

SergeyTravin:


SergeyTravin:
@pentaho-lmartins, @mbatchelor please review it is fix for regression caused by 
https://github.com/pentaho/pentaho-platform-plugin-dashboards/commit/d203a719a22c2cd8c2e8f6b435aa22d3de1c9127

PR #621, opened 2018-04-17
[PDB-2002] Dashboard with Refresh Interval activated containing a PRP…

ppatricio:
…T shows refreshing on top of other PUC Tabs

@tmorgner  @diogofscmariano @pentaho-lmartins 

showLoadingIndicator logic at report viewer is causing a loading screen all over PUC at every load and report refresh.

### DIFF
--- a/core/src/main/javascript/reportviewer/report.html
+++ b/core/src/main/javascript/reportviewer/report.html
@@ -54,15 +54,7 @@
 
       }
     };
-
-
-    try {
-      if(window.top.showLoadingIndicator) {
-        window.top.showLoadingIndicator();
-      } else if(window.parent.showLoadingIndicator) {
-        window.parent.showLoadingIndicator();
-      }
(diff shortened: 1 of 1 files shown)


diogofscmariano:
Don't we want to ignore the code you are removing in case the Dashboard has a refresh interval activated, but keep it active for the other cases? Or it is always enabled?

ppatricio:
It's only enabled at Dashboard context. Every report loading /refreshing in a Dashboard context is calling this loading PUC dialog.
Previous to this PR we had two different behaviors:
-loading /refreshing a report in PUC- did not call the loading screen in report.html
-loading /refreshing a report inside a dashboard in PUC- call the loading screen in report.html
Removing that logic solved this difference that was in fact the cause to the disruptive behavior reported with refresh intervals with multiple PUC tabs.

diogofscmariano:
All good

PR #622, opened 2018-04-24
[SP-4270][BISERVER-13895]-Cannot read property 'fireOutputParam' of u…

AliaksandrShuhayeu:
…ndefined when


AliaksandrShuhayeu:
@pentaho-lmartins, @mchen-len-son, could you please take a look?

tmorgner:
Same as #620 therefore merged.

PR #623, opened 2018-04-24
[SP-4269][BISERVER-13895]-Cannot read property 'fireOutputParam' of undefined when

AliaksandrShuhayeu:
[SP-4269][BISERVER-13895]-Cannot read property 'fireOutputParam' of undefined when
@pentaho-lmartins, @mchen-len-son, could you please take a look?


tmorgner:
Seems Wingman died while waiting for this request to be validated. But given that it is identical to the other pull requests of the series, it should be safe. (for reference: #620 contains the master fix)

PR #624, opened 2018-04-26
[PRD-5982] Cascading parameters are not reloaded on 7.0, 7.1 and 8.0 …

ricardosilva88:
…if using bean shell script

@tmorgner 
Applied your refactoring and tested - https://github.com/tmorgner/pentaho-platform-plugin-reporting/commit/81dbce54bfbabd8853f6c752265b5338d5eb05aa

Created unit resting for ParameterDependencyGraph


tmorgner:
Thank you. Wow, even Wingman is happy. Now that must be a first! :)

PR #627, opened 2018-05-21

cravobranco:

   * Added client side support to handle reports which date is calculated based on client timezone
   * Refactored the date parser based on timezone type value, which can be one of the following:
    1) server - timezone already handled by the server
    2) client - timezone handled in the client side (see pentaho-platform-plugin-common-ui - formatting.js)
    3) utc - timezone forced to UTC
    4) customized timezone - timezone forced to customized timezone

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/ReportContentUtil.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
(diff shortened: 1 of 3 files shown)


cravobranco:
@pentaho-lmartins , @pamval , @mbatchelor , @mchen-len-son , @tmorgner  please review and merge if it's all ok

Notes: This should be reviewed together with:
https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1296

Main changes:
- Added client side support to handle reports which date is calculated based on client timezone
- Refactored the date parser based on timezone type value, which can be one of the following:
1) server - timezone already handled by the server
2) client - timezone handled in the client side (see pentaho-platform-plugin-common-ui - formatting.js)
3) utc - timezone forced to UTC
4) customized timezone - timezone forced to customized timezone

cravobranco:
This case is now handled in #636

PR #631, opened 2018-06-13
[PPP-4108] Use of vulnerable component jackson-databind-2.8.8.jar, ja…

pentaho-lmartins:
…ckson-databind-2.9.2.jar, jackson-mapper-asl-1.9.2.jar CVE-2017-7525, CVE-2017-15095, CVE-2017-15095

This is one of a series of PRs. See further details in https://github.com/pentaho/maven-parent-poms/pull/66.

**Do not merge without checking the status of the JIRA case.**

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -92,15 +92,11 @@
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-core</artifactId>
+      <artifactId>jackson-databind</artifactId>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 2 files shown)


core/pom.xml
```diff
@@ -92,16 +92,12 @@
...
-      <artifactId>jackson-core</artifactId>
+      <artifactId>jackson-databind</artifactId>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
```
graimundo: From a diff perspective it would be easier to understand the change if you deleted the jackson-core artifact instead of deleting the jackson-databind artifact and renaming the jackson-core to jackson-databind.

PR #636, opened 2018-08-03

cravobranco:

* As a workaround, when there is only date configured (and no time) then we parse only as date to avoid errors caused by change of day due to client/server different timezones (because reporting mechanism assumes no time = 00:00:00)

Note: a whole refactor should be done to integrate new Java 8 DateTime API and then this workaround can be removed.


cravobranco:
@pentaho-lmartins , @tmorgner , @pamval , @mbatchelor Please review and merge if it's all ok!

Changes description:
- As a workaround when there is only date configured (and no time), then we parse only as date to avoid errors caused by change of day due to client/server different timezones (because reporting mechanism assumes no time = 00:00:00).
- A whole refactor should be done to integrate new Java 8 DateTime API and then this workaround can be removed.

PR #639, opened 2018-08-20
[BISERVER-13775] PUC-Unable to generate report in Excel

LeonardoCoelho71950:
This wasn't actually a bug. In Excel 97, the maximum number of allowed rows is 65536. If the file is bigger, it is impossible to generate a proper excel file.  Before the ReportProcessingException was silenced during execution, but now the exception is thrown. Moreover, to give a more descriptive error message, all cascading exceptions' error messages are included.  

Instead of: "Request failed"
The error message is:
 " Failed to process the report.
    Other failure.
    Invalid row number (65536) outside allowable range (0..65535).
 "

@pamval @duarteteixeira 
@pentaho-lmartins @ricardosilva88

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+ * Copyright (c) 2002-2018 Hitachi Vantara..  All rights reserved.
  */
 
 package org.pentaho.reporting.platform.plugin;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java
```diff
@@ -970,6 +971,9 @@ public boolean execute() throws Exception {
       log.info( "Report execution interrupted." );
     } catch ( Exception e ) {
       log.error( Messages.getInstance().getString( "ReportPlugin.executionFailed" ), e ); //$NON-NLS-1$
+      if ( e instanceof ReportProcessingException ) {
```
pamval: Why not a specific catch clause for ReportProcessingException?

LeonardoCoelho71950: I could specify a catch clause for ReportProcessingException before the Exception's catch clause. 
But I would need to repeat line 973 inside the new catch.  https://github.com/pentaho/pentaho-platform-plugin-reporting/blob/b59e2f373e06b65e73614cd86e220e01633062be/core/src/main/java/org/pentaho/reporting/platform/plugin/SimpleReportingComponent.java#L973

Please tell me what version do you prefer, @pamval ?

PR #643, opened 2018-09-26
[PRD-2034] - Dashboard refresh error "pentaho-ajax.js.pentahoResponse…

ricardosilva88:
…(): TypeError: Cannot read property 'className' of null"

During intense report refresh, with times between 1 and 2 seconds time to refresh (when inside a dashboard), domClass may become null for a fraction of time, which causes the error - "pentaho-ajax.js.pentahoResponse(): TypeError: Cannot read property 'className' of null"


ricardosilva88:
Please note that commit message is wrong - more details on https://jira.pentaho.com/browse/PDB-2034

PR #644, opened 2018-10-01
[PIR-1414] Report does not render complete data when Query Limit, Des…

ssamora:
…ign Query limit are set and exported via Schedule/Run in Background
@pentaho-lmartins @tmorgner


tmorgner:
Looks good. A sweet and simple fix. Merged.

PR #673, opened 2019-01-03
[BISERVER-14069] - Fixed Language Translation Properties Files Not Pi…

RPAraujo:
…cked up from PRPTs when running on Server

@pentaho/tatooine @tmorgner please review

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2018 Hitachi Vantara.  All rights reserved.
+ * Copyright (c) 2002-2019 Hitachi Vantara.  All rights reserved.
  */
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/pentaho/reporting/platform/plugin/output/CachingPageableHTMLOutput.java
```diff
@@ -390,6 +390,7 @@ private Serializable computeCacheKey( final MasterReport report ) throws BeanExc
       sourceKey = computeDefSourceKey( report, definitionSource );
     }
     final HashMap<String, String> params = new HashMap<>();
+    params.put("report-locale", report.getReportEnvironment().getLocale().toString() );
```
pamval: Check style issue, please fix

tmorgner: Adding the locale information into the same collection as the report's parameters will clash the moment a user has a report parameter named "report-locale". Add that information as new element into the "key" array list below. Remember: The only requirement of the result of this function is that it is serializable and that it uniquely identifies a report with all its configuration details and parameters. So adding the locale string down there is safe and a lot better than risking namespace clashes.

RPAraujo: Done

RPAraujo: Done

PR #689, opened 2019-02-27
[BISERVER-13851] - Checks if the paramDefn is initialized before trying to access it.

andrefrua:


andrefrua:
@pentaho/bobafett Please review

PR #697, opened 2019-03-18
[BACKLOG-28114] Improve reports loading speed

nantunes:
@pentaho/millenniumfalcon please review.

I would not squash the commits, they handle two separate issues.


nantunes:
@dcleao updated.

PR #707, opened 2019-04-05
[PDI-17667] Use of vulnerable component jetty 8.1.15, CVE-2017-9735, …

smmribeiro:
…CVE-2017-7657 CVE-2015-2080

[PPP-4183] Use of Vulnerable Component: Multiple Jetty Components [Listed in Description] (CVE-2017-7656 | CVE-2017-7657 | CVE-2017-7658 | CVE-2015-2080 | CVE-2017-9735 )

Depends on pentaho/maven-parent-poms#121

@pentaho-lmartins @pentaho/tatooine
@graimundo @nantunes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,10 +112,6 @@
             <artifactId>*</artifactId>
             <groupId>com.sun.jersey</groupId>
           </exclusion>
-          <exclusion>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
-          </exclusion>
         </exclusions>
       </dependency>
       <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -112,6 +112,10 @@
             <artifactId>*</artifactId>
             <groupId>com.sun.jersey</groupId>
           </exclusion>
+          <exclusion>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+          </exclusion>
         </exclusions>
       </dependency>
       <dependency>


pom.xml
```diff
@@ -112,6 +112,10 @@
             <artifactId>*</artifactId>
             <groupId>com.sun.jersey</groupId>
           </exclusion>
+          <exclusion>
+            <groupId>javax.servlet</groupId>
```
graimundo: @smmribeiro why is this needed? With what does it conflict?

smmribeiro: @graimundo: Interestingly, this repository only started to have problems with the Jetty upgrade after the https://github.com/pentaho/pentaho-platform-plugin-reporting/commit/8a819a6c547c7d04fa3bdb7918f6c4b3f06bac70 commit!

After that commit, a javax.servlet-api artifact started to appear as transitive and made the build fail.

I've done a little research and found out that, until version 3.x, the artifactId for servlet-api was **"servlet-api"** and after that it was changed to **"javax.servlet-api"** (you can see that in the pom.xml as the two forms exist)... I think that it may be the reason why the transitive dependency "passes on" now.

The class that shows errors while building without this change is [core] org.pentaho.reporting.platform.plugin.connection.CdaPluginLocalQueryBackend.

That's why I had to add this exclusion.

smmribeiro: Interestingly, seems that with the changes operated on the other repositories, this repository works again without changes :-)
Resolving this (and probably close the PR).

smmribeiro:
Closing PR: to simplify Karaf's upgrade a different approach was defined!

PR #738, opened 2019-09-24
[BISERVER-14306] Multi Value List parameter values are reset after changing output type in a prpt report

gryphendowre:
@pentaho/tatooine @ssamora @pentaho-lmartins @tmorgner

* [BISERVER-14306] Added extra functionality and reversed logic for ParameterDependencyGraph
* [BISERVER-14306] Changed boolean logic to check for dependencies
instead of availability and in ParameterDependencyGraph checked for
instance of TableDataFactory to return an empty table instead of null
* [BISERVER-14306] Added extra check for getSelections
* [BISERVER-14306] Updated tests
* [BISERVER-14306] Updated client javascript to ensure the selected
value is set properly when switching date objects (this is generally
working, but for those date objects that have default values, this loses
it's value)
* [BISERVER-14306] Cleanup of javascript and java code

Pull Request is from a set of two PRs:
* https://github.com/pentaho/pentaho-platform-plugin-reporting/pull/738
* https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1475

### DIFF
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
@@ -89,10 +89,6 @@ public class ParameterDependencyGraph {
     this.allParametersProcessed = allParametersProcessed;
   }
 
-  public boolean getAllParametersProcessed() {
-    return this.allParametersProcessed;
-  }
-
   public LinkedHashMap<String, Set<String>> getDependencyGraph() {
     return dependencyGraph;
   }
@@ -140,7 +136,7 @@ public class ParameterDependencyGraph {
   }
 
   /**
-   * Stream and verify that the parameter exists as a dependency in the dependency graph
+   * Verify that the parameter exists as a dependency in the dependency graph
    * @param parameterName
    * @return
    */
@@ -242,6 +238,8 @@ public class ParameterDependencyGraph {
     }
 
     final DataFactory dataFactoryForQuery = cdf.getDataFactoryForQuery( queryName );
+    // We are also checking for TableDataFactory, because this Data Factory is a manual table that cannot
+    // have or be used in a dependency query. So it's list should be empty.
     if ( dataFactoryForQuery == null || dataFactoryForQuery instanceof TableDataFactory ) {
       return Collections.emptyList();
     }
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
@@ -52,15 +52,15 @@ import java.util.stream.Collectors;
  * this would return a mapping "City -> [Country]".
  * <p>
  * If there is a problem with computing dependency information or if the queries use a datasource
- * that cannot provide dependency information, this implementation will return "true" on
- * "noDependencyInformationAvailable"
+ * that cannot provide dependency information, this implementation will return "false" on
+ * "allParametersProcessed"
  * and will return all other parameters as dependent values (to indicate that any change may be a
  * cause of change in the parameter values).
  */
 public class ParameterDependencyGraph {
   private static final String SYS_IGNORE_PARAM = "::org.pentaho.reporting";
 
-  private boolean dependencyInformationAvailable = true;
+  private boolean allParametersProcessed;
   private LinkedHashMap<String, Set<String>> dependencyGraph;
   private Set<String> allParameterNames;
 
@@ -70,7 +70,8 @@ public class ParameterDependencyGraph {
                                    final Map<String, Object> computedParameterValues ) {
     this.dependencyGraph = new LinkedHashMap<>();
     this.allParameterNames = new HashSet<>( reportParameter.keySet() );
-    processDependentParameters( report, reportParameter, parameterContext, new StaticDataRow( computedParameterValues ) );
+    this.allParametersProcessed = processDependentParameters( report, reportParameter, parameterContext,
+      new StaticDataRow( computedParameterValues ) );
   }
 
   /**
@@ -80,15 +81,16 @@ public class ParameterDependencyGraph {
    */
   ParameterDependencyGraph( String... allParameterNames ) {
     this.dependencyGraph = new LinkedHashMap<>();
+    this.allParametersProcessed = true;
     this.allParameterNames = new HashSet<>( Arrays.asList( allParameterNames ) );
   }
 
-  public void setDependencyInformationAvailable( boolean dependencyInformationAvailable ) {
-    this.dependencyInformationAvailable = dependencyInformationAvailable;
+  void setAllParametersProcessed( boolean allParametersProcessed ) {
+    this.allParametersProcessed = allParametersProcessed;
   }
 
-  public boolean getDependencyInformationAvailable() {
-    return this.dependencyInformationAvailable;
+  public boolean getAllParametersProcessed() {
+    return this.allParametersProcessed;
   }
 
   public LinkedHashMap<String, Set<String>> getDependencyGraph() {
@@ -108,7 +110,7 @@ public class ParameterDependencyGraph {
   }
 
   public Set<String> getDependentParameterFor( String parameterName ) {
-    if ( !dependencyInformationAvailable ) {
+    if ( !allParametersProcessed ) {
       return Collections.emptySet();
     }
 
@@ -121,7 +123,7 @@ public class ParameterDependencyGraph {
   }
 
   public Set<String> getAllDependencies( Iterable<String> parameterNames ) {
-    if ( !dependencyInformationAvailable ) {
+    if ( !allParametersProcessed ) {
       return Collections.unmodifiableSet( allParameterNames );
     }
 
@@ -190,13 +192,13 @@ public class ParameterDependencyGraph {
                                               ParameterContext parameterContext,
                                               DataRow parameterValues ) {
 
+    boolean isDependencyInfoAvail = true;
     try {
       final DesignTimeDataFactoryContext factoryContext = new DesignTimeDataFactoryContext( report );
       final CompoundDataFactory cdf = CompoundDataFactory.normalize( report.getDataFactory() );
       final CompoundDataFactory derive = (CompoundDataFactory) cdf.derive();
       derive.initialize( factoryContext );
 
-      boolean anyDependencyProcessed = false;
       try {
         for ( final ParameterDefinitionEntry entry : reportParameter.values() ) {
           final List<String> dependentParameter = computeNormalLineage( parameterContext, entry );
@@ -204,28 +206,25 @@ public class ParameterDependencyGraph {
 
           // No dependency information at all - No need to continue
           if ( queryDependencies == null ) {
-            dependencyInformationAvailable = false;
+            isDependencyInfoAvail = false;
             continue;
           }
 
           if ( dependentParameter != null && dependentParameter.size() > 0 ) {
             dependentParameter.forEach( p -> addDependency( p, entry.getName() ) );
-            anyDependencyProcessed = true;
           }
           if ( queryDependencies != null && queryDependencies.size() > 0 ) {
             queryDependencies.forEach( p -> addDependency( p, entry.getName() ) );
-            anyDependencyProcessed = true;
           }
         }
       } finally {
         derive.close();
       }
-
-      return anyDependencyProcessed;
     } catch ( ReportDataFactoryException re ) {
       DebugLog.log( "Failed to compute dependency information", re );
-      return false;
+      isDependencyInfoAvail = false;
     }
+    return isDependencyInfoAvail;
   }
 
   private List<String> computeListParameterLineage( CompoundDataFactory cdf,
@@ -260,8 +259,8 @@ public class ParameterDependencyGraph {
     }
   }
 
-  public boolean isDependencyInformationAvailable() {
-    return dependencyInformationAvailable;
+  public boolean areAllParametersProcessed() {
+    return allParametersProcessed;
   }
 
   private static String extractFormula( final ParameterContext parameterContext,
--- a/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
+++ b/core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
@@ -23,6 +23,7 @@ import org.pentaho.reporting.engine.classic.core.DataRow;
 import org.pentaho.reporting.engine.classic.core.MasterReport;
 import org.pentaho.reporting.engine.classic.core.ReportDataFactoryException;
 import org.pentaho.reporting.engine.classic.core.StaticDataRow;
+import org.pentaho.reporting.engine.classic.core.TableDataFactory;
 import org.pentaho.reporting.engine.classic.core.designtime.datafactory.DesignTimeDataFactoryContext;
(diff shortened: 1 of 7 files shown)


buildguy:
### :gear: Build finished in 5m 9s with command:
```
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd -Daudit -P !assemblies -pl core
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|602|0|3|

|[Details](http://wingman2.pentaho.net/job/projects/job/platform-reporting-plugin/job/PR-738/1/testReport)|
|----|

:information_source:  _This is an automatic message_

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
```diff
@@ -69,29 +70,25 @@ public ParameterDependencyGraph( final MasterReport report,
...
-  ParameterDependencyGraph( boolean noDeps, String... allParameterNames ) {
+  ParameterDependencyGraph( String... allParameterNames ) {
     this.dependencyGraph = new LinkedHashMap<>();
     this.allParameterNames = new HashSet<>( Arrays.asList( allParameterNames ) );
-    this.noDependencyInformationAvailable = noDeps;
   }
 
-  public void setNoDependencyInformationAvailable( boolean noDependencyInformationAvailable ) {
-    this.noDependencyInformationAvailable = noDependencyInformationAvailable;
+  public void setDependencyInformationAvailable( boolean dependencyInformationAvailable ) {
```
tmorgner: You might want to make this setter package local. Public stuff is callable from everywhere, and most importantly has strict rules on when we may change or remove it. So its better to not make things public unless you really really need it to be public. As this method is only used in the unit test, which sits in the same package, package local will do fine without carrying the burden of those wacky rules around deprecation.

gryphendowre: Agreed, will make this change.

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
```diff
@@ -187,15 +196,16 @@ private boolean processDependentParameters( MasterReport report,
...
+      boolean anyDependencyProcessed = false;
       try {
         for ( final ParameterDefinitionEntry entry : reportParameter.values() ) {
           final List<String> dependentParameter = computeNormalLineage( parameterContext, entry );
           final List<String> queryDependencies = computeListParameterLineage( derive, entry, parameterValues );
 
           // No dependency information at all - No need to continue
           if ( queryDependencies == null ) {
-            return false;
+            dependencyInformationAvailable = false;
```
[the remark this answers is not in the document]
gryphendowre: Agreed, we may need to rename the variable altogether as well. Recommending changing it to allParameterDependenciesProcessed or allParametersProcessed.  That way, the semantics of it still state that either all the parameters were successfully processed or not (due to dependency query issues or exceptions).  Because we can have no dependencies and it not be an issue. The previous logic conflated No Dependencies and Dependencies Processing Issue together, which causes issues with some use cases in the report.

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterXmlContentHandler.java
```diff
@@ -747,7 +752,7 @@ Element createParameterElement( final ParameterDefinitionEntry parameter,
 
         //parameter dependencies: see backlog-7980
         boolean shouldValidateOnServer =
-          !dependencies.isNoDependencyInformationAvailable()
+          !dependencies.getDependencyGraph().isEmpty()
```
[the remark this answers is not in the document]
gryphendowre: I tried different tests with the 11 different tickets I am verifying against (BISERVER-14306 BISERVER-14206 BISERVER-14297 BISERVER-14315 BISERVER-14308 PRD-6082 BISERVER-13435 PRD-5982 BACKLOG-26181 BISERVER-14198 BISERVER-14254) and I came to the conclusion that "isNoDependencyInformationAvailable / isDependencyInformationAvailable / dependencies.getDependencyGraph().isEmpty()" seems to be not helping.  By having one of these checks in, it keeps BACKLOG-26181 from working (parameters are hidden). By having an && check in instead, it keeps the bean shell script from running on PRD-5982 properly (it'll update only once and never again). When I remove that check and only check against "shouldAlwaysValidateOnServer", then all the tickets respond positively to it.  My initial change to !dependencies.getDependencyGraph().isEmpty() was expecting that if any dependencies existed, then it should be validated.  It actually should only be validated if the parameter has a dependency, which is checked in the next conditional statement, so my initial change was unnecessary.

The original code was conflating No Dependencies and Unable to Process Dependencies before, by reversing my logic I'm now stating that you are able to process dependencies successfully (regardless of if there are dependencies or not). So in essence, the old code check was verifying if dependencies existed.  But this is basically covered by checking if a parameter has a dependency in the next conditional statement, which then tells the client code to validate on the server.  So by removing the check, we end up having less redundancy.

There were some issues as well with the old code and using the isDependencyInformationAvailable, similar problems, where if the old one came back overall with true, or if isDependencyInformationAvailable was true, then it would state every parameter should be validated (which is not true for every parameter) and caused some issues with the reporting plugin.  We only need to force validation of parameters that need it, so by removing the redundant and too general boolean, we end up being more accurate (the 11 tickets are satisfied).

buildguy:
### :gear: Build finished in 4m 11s with command:
```
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd -Daudit -P !assemblies -pl core
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|602|0|3|

|[Details](http://wingman2.pentaho.net/job/projects/job/platform-reporting-plugin/job/PR-738/2/testReport)|
|----|

:information_source:  _This is an automatic message_

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
```diff
@@ -233,7 +243,7 @@ private boolean processDependentParameters( MasterReport report,
     }
 
     final DataFactory dataFactoryForQuery = cdf.getDataFactoryForQuery( queryName );
-    if ( dataFactoryForQuery == null ) {
+    if ( dataFactoryForQuery == null || dataFactoryForQuery instanceof TableDataFactory ) {
```
dcleao: I'm sure this should be obvious, however, it looks surprising for an outsider. What is it with this type of data factory which makes us immediately return with an empty list? Please, consider adding a comment.

buildguy:
### :gear: Build finished in 4m 7s with command:
```
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd -Daudit -P !assemblies -pl core
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|602|0|3|

|[Details](http://wingman2.pentaho.net/job/projects/job/platform-reporting-plugin/job/PR-738/3/testReport)|
|----|

:information_source:  _This is an automatic message_

core/src/main/javascript/reportviewer/reportviewer-prompt.js
```diff
@@ -652,7 +652,11 @@ define([
           var counter = 0;
           for (var i in param.values) {
             var paramVal = param.values[i];
-            if (value.indexOf(paramVal.value) > -1) {
+            if (value.indexOf(paramVal.value) > -1
+              //Datepickers require some extra care (this is needed if the values list has a stored default value and
+              // keeps it ready and selected=false.)
+              || (param.attributes['parameter-render-type'] === 'datepicker'
+                && this._compareDatesOnly(paramVal.value, value[0], param.timezoneHint, this._createFormatter(param)))) {
```
dcleao: Shouldn't you be comparing each position of the `value` array with `paramVal.value`? You're only comparing the first position.
Maybe your comment explains this, but I couldn't understand it.

core/src/main/javascript/reportviewer/reportviewer-prompt.js
```diff
@@ -715,7 +719,6 @@ define([
 
             //Timezone is not found in a value, check the hint
             if (timezoneHint) {
```
dcleao: Why is the below replace not needed anymore (e.g from `\+xxx` to `+xxx`)?

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
```diff
@@ -140,6 +139,18 @@ public void setAllParameterNames( Set<String> allParameterNames ) {
     return Collections.unmodifiableSet( dependencyGraph.keySet() );
   }
 
+  /**
+   * Stream and verify that the parameter exists as a dependency in the dependency graph
```
dcleao: "Stream and verify" - streaming is about the particular implementation, no? Don't think it needs to surface in the description.

core/src/test/javascript/reportviewer-prompt-spec.js
```diff
@@ -408,12 +413,49 @@ define(["reportviewer/reportviewer-prompt", "reportviewer/reportviewer-logging",
...
+            ],
+            attributes: {'parameter-render-type': 'list'}
           };
           reportPrompt._applyUserInput("anyname", ["test", "test1", "test2"], parameter);
           expect(parameter.forceUpdate).toBe(true);
         });
 
+        // [BISERVER-14306]
+        // These 2 tests only occur in code when there's a default value for a Date object.  The selection goes through
+        // A loop to first remove the bad date and then adds the new date value. Code was updated to handle the forced
```
dcleao: "A default value for a Date object" -> for a date parameter?
"remove the bad date" -> what is a bad date?

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
```diff
@@ -140,6 +139,18 @@ public void setAllParameterNames( Set<String> allParameterNames ) {
...
+   * @param parameterName
+   * @return
+   */
+  public boolean doesDependencyExist( String parameterName ) {
+    return dependencyGraph
+          .values()
+          .stream()
+          .flatMap( Set::stream )
+          .anyMatch( s -> s.compareTo( parameterName ) == 0 );
```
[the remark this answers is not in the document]
gryphendowre: It is, it's comparing if a string value in the flatMap is equal to the parameter name. The goal is to make sure we're checking for existence that the parameter is itself a dependency of another, a reverse of the getDependencyFor which looks for dependencies of a parameter.

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
```diff
@@ -250,8 +260,8 @@ private boolean processDependentParameters( MasterReport report,
     }
   }
 
-  public boolean isNoDependencyInformationAvailable() {
-    return noDependencyInformationAvailable;
+  public boolean areAllParametersProcessed() {
```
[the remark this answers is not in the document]
gryphendowre: Will get rid of, looks like the old method was not being used either

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
gryphendowre: Comment added.

core/src/main/javascript/reportviewer/reportviewer-prompt.js
gryphendowre: Further comment added. In the case where that checks occurs, the value object is a selected date, but is wrapped as an array with 1 object. So we only need to check the first item to get the selected value. The paramVal.val comes from the array of currently selected/default values and that is looped properly.

core/src/main/javascript/reportviewer/reportviewer-prompt.js
gryphendowre: The Timezone hint is calculated on the server side and we removed the backslash in the java code. The check was only necessary if we were initially calculating it on the server side as \+xxx.
The original code check here was also never working, as the regex was not a proper format and always left it as \+xxx, then when it went to compare dates, it would always return "true" for an invalid value and never update. By fixing it initially and then removing it, we uncovered other issues that required other javascript fixes (as seen in this file too).

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
gryphendowre: Will remove stream, leave verify

core/src/test/javascript/reportviewer-prompt-spec.js
gryphendowre: In the Report Designer, you can have a calculated default date value by a query. In the javascript, it formats this as a list of date objects (label/value/attributes). With the calculated date default as the first item in the list. So these test cases check this use case specifically.
Bad date not appropriate, updating comment to say the previously selected date

buildguy:
### :gear: Build finished in 4m 7s with command:
```
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd -Daudit -P !assemblies -pl core
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|602|0|3|

|[Details](http://wingman2.pentaho.net/job/projects/job/platform-reporting-plugin/job/PR-738/4/testReport)|
|----|

:information_source:  _This is an automatic message_

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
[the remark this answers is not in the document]
dcleao: I got the idea of the method. I just don't understand why you're using the `compareTo` method, instead of the simpler `equals` method.

core/src/main/javascript/reportviewer/reportviewer-prompt.js
dcleao: Oh, I see now the reason why you only need to check the first position - the data picker does not allow multiple selection.

core/src/main/javascript/reportviewer/reportviewer-prompt.js
dcleao: Yes, it looks like it was a missing a slash to make it work `/\\\+/`. I missed that on the first read.
Anyway, I didn't see any server-side code removal related with timezone hints and thus my doubt.
I assume the removal was made in the context of another issue.
Thanks!

diogofscmariano:
No comments in this change

core/src/main/java/org/pentaho/reporting/platform/plugin/ParameterDependencyGraph.java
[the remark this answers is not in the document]
gryphendowre: Good point, just blanked I guess. I'll apply it in the next commit, want to wait for any other comments before committing any more code.

buildguy:
### :gear: Build finished in 4m 18s with command:
```
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd -Daudit -P !assemblies -pl core
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|602|0|3|

|[Details](http://wingman2.pentaho.net/job/projects/job/platform-reporting-plugin/job/PR-738/5/testReport)|
|----|

:information_source:  _This is an automatic message_

PR #819, opened 2023-11-15
[BACKLOG-39139] Fix css overflow behavior to prevent content from being

peterrinehart:
cut off when rendered in an iframe.

### DIFF
--- a/core/src/main/javascript/reportviewer/report.html
+++ b/core/src/main/javascript/reportviewer/report.html
@@ -36,7 +36,10 @@
       disableFlashStorage: true
     };
 
-
+    var inSchedulerDialog = window.location.pathname.toLowerCase().indexOf('parameterui') >= 0;
+    if (inSchedulerDialog) {
+      document.documentElement.style.setProperty( '--var-default-overflow', "visible" );
+    }
 
     /**
      * This is called when a gwt module has been loaded. We depend on these so we'll defer loading of functionality until
--- a/core/src/main/javascript/reportviewer/reportviewer.css
+++ b/core/src/main/javascript/reportviewer/reportviewer.css
@@ -1,5 +1,9 @@
+:root {
+  --var-default-overflow: hidden;
+}
+
 * {
-	overflow: hidden;
+	overflow: var( --var-default-overflow );
 }
 
 html, body {


buildguy:
### :white_check_mark: Build finished in 11m 13s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl core
```
---
:ok_hand: All tests passed!
 
Tests run: **540**, Failures: **0**, Skipped: **3** &nbsp;&nbsp; [Test Results](https://wingman.orl.eng.hitachivantara.com/job/projects/job/platform-reporting-plugin/job/PR-819/1/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

ddiroma (approved):

core/src/main/javascript/reportviewer/report.html
```diff
@@ -36,7 +36,10 @@
       disableFlashStorage: true
     };
 
-
+    var inSchedulerDialog = window.location.pathname.toLowerCase().indexOf('parameterui') >= 0;
```
dcleao, line 39: The existing responsive parameter UI contract involves two parties.

One in the schedule parameters dialog, the host. It has the following styles applied:
https://github.com/pentaho/pentaho-commons-gwt-modules/blob/3d744604b2d1c211ce84faa341f59b074ef0e87c/widgets/src/main/resources/org/pentaho/gwt/widgets/themes/public/themes/ruby/globalRuby.css#L2064-L2083

The other party is the HTML document of the parameter UI for a certain type of scheduled content, which is displayed in an iframe in the host party. A parameter UI document that opts in to responsive mode (for responsive themes) must specify the classes `responsive` and `parameter-ui` on its `body` element. The following styles will apply, on responsive themes (atm, only the ruby theme):
https://github.com/pentaho/pentaho-commons-gwt-modules/blob/3d744604b2d1c211ce84faa341f59b074ef0e87c/widgets/src/main/resources/org/pentaho/gwt/widgets/themes/public/themes/ruby/globalRuby.css#L1942C5-L1946

The change being done here likely does not achieve the same level of responsiveness, given the host `iframe` will always have the same size, and, ideally, there should be a single vertical bar displayed on the host document, and not inside the `iframe`. Also, this change affect all themes.

This change is acceptable, compared to using the "responsive parameter UI contract" presented above, given that the contained prompt parameter components, and the prompt`s overall layout, are not prepared to behave responsively, yet. And that will take more work.

PR #848, opened 2024-04-30
[BACKLOG-39583] JDK17 - Build issue for https://github.com/pentaho/pe…

eddie-martinez:
…ntaho-platform-plugin-reporting.git

### DIFF
--- a/core/src/test/java/org/pentaho/reporting/platform/plugin/JobManagerContextTest.java
+++ b/core/src/test/java/org/pentaho/reporting/platform/plugin/JobManagerContextTest.java
@@ -34,14 +34,14 @@ import org.pentaho.reporting.platform.plugin.async.AsyncExecutionStatus;
 import org.pentaho.reporting.platform.plugin.async.IAsyncReportState;
 import org.pentaho.reporting.platform.plugin.async.IPentahoAsyncExecutor;
 
-
 import java.io.IOException;
 import java.util.UUID;
 
 import static junit.framework.TestCase.assertFalse;
 import static junit.framework.TestCase.assertTrue;
-import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/pentaho/reporting/platform/plugin/JobManagerContextTest.java
```diff
@@ -53,15 +52,23 @@ public class JobManagerContextTest {
...
+    mockStaticPentahoSessionHolder = mockStatic( PentahoSessionHolder.class );
     session = mock( IPentahoSession.class );
     uuid = UUID.randomUUID();
     final IPentahoAsyncExecutor executor = mock( IPentahoAsyncExecutor.class );
     state = mock( IAsyncReportState.class );
     report = mock( MasterReport.class );
     when( executor.getReportState( uuid, session ) ).thenReturn( state );
     PentahoSystem.registerObject( executor, IPentahoAsyncExecutor.class );
+    mockStaticReportCreator.when(  () -> ReportCreator.createReportByName( any() ) ).thenReturn( report );
+    mockStaticPentahoSessionHolder.when( PentahoSessionHolder::getSession ).thenReturn( session );
```
peterrinehart, line 71: Why did all of these get moved to the setup method?  They're far better off where they were in the try-with-resources block to ensure they don't interfere with any other tests.

peterrinehart (changes requested):
See comments, also update the copyright headers in the test classes updated.

core/src/test/java/org/pentaho/reporting/platform/plugin/JobManagerTest.java
```diff
@@ -93,37 +98,42 @@ public class JobManagerTest {
...
+  }
+
   @AfterClass
   public static void tearDown() throws Exception {
     PentahoSystem.shutdown();
   }
 
   @Before
   public void before() {
+    pentahoSessionHolderMockedStatic = mockStatic( PentahoSessionHolder.class );
```
peterrinehart, line 129: We need to have a good reason for initializing this here instead of in a try-with-resources block in the test method.

